PR #3, opened 2017-01-07
Feat/idea

IDragonfire:
It is not a good idea to share the complete .idea folder ...
I have different files locally and on each branch switch stuff is mixed up ...


micheljung:
Please don't use `feat` prefixes for branches, stick to the git flow conventions.

IDragonfire:
This is only to show you what happen if another user import the project ...

PR #8, opened 2017-01-09
You can now enable PATCH Requests for some Entities, just add the `@U…

IDragonfire:
You can now enable PATCH Requests for some Entities, just add the `@UpdatePermission(expression = "is logged in")` annotation


src/main/java/com/faforever/api/config/security/WebSecurityConfig.java
```diff
@@ -47,7 +47,9 @@ public AuthenticationManager authenticationManagerBean() throws Exception {
   @Override
   protected void configure(HttpSecurity http) throws Exception {
     // @formatter:off
-    http.headers()
+      http
+      .csrf().disable() // http://stackoverflow.com/a/29917946
```
micheljung: This disables CSRF globally, but it should be disabled for the login form

src/main/java/com/faforever/api/data/package-info.java
```diff
@@ -1,4 +1,15 @@
 /**
  * Contains classes to access data according to the <a href="http://jsonapi.org/">JSON-API specification</a>.
  */
+// everybody can read from the api
```
micheljung: **E**verybody

src/main/java/com/faforever/api/data/package-info.java
```diff
@@ -1,4 +1,15 @@
...
+@ReadPermission(any={Role.ALL.class })
+// By default restrict all data manipulation operation
+// @UpdatePermission(expression = "is logged in")
```
micheljung: Why is this commented out?

src/main/java/com/faforever/api/config/elide/checks/IsUser.java
```diff
@@ -0,0 +1,14 @@
+package com.faforever.api.config.elide.checks;
+
+import com.yahoo.elide.security.User;
+import com.yahoo.elide.security.checks.UserCheck;
+
+/**
+ * @author Dragonfire
+ */
```
micheljung: Remove

src/main/java/com/faforever/api/config/elide/checks/IsUser.java
```diff
@@ -0,0 +1,14 @@
...
+public class IsUser extends UserCheck {
```
micheljung: Would `IsAuthenticated` be more descriptive? Because there can also be anonymous users.

src/main/java/com/faforever/api/config/elide/checks/IsUser.java
```diff
@@ -0,0 +1,14 @@
...
+  @Override
+  public boolean ok(User user) {
+    return user != null & user.getOpaqueUser() != null;
+  }
+}
```
micheljung: The structure of this class is not consistent with the existing classes and as recommended by Elide: http://elide.io/pages/guide/03-security.html

src/main/java/com/faforever/api/config/security/WebSecurityConfig.java
IDragonfire: `When should you use CSRF protection? Our recommendation is to use CSRF protection for any request that could be processed by a browser by normal users. If you are only creating a service that is used by non-browser clients, you will likely want to disable CSRF protection.`

PR #44, opened 2017-02-08
Add test for /clans/ + Error Reporting

IDragonfire:
* More clan tests
* Clan Integration tests
* Fix clan tests (this also fixed travis)
* Introduce global exception handling
* Clan Name/Tag mus be now unique

Clanapp is now feature complete: https://github.com/FAForever/clans/issues/36

### DIFF
--- a/src/main/java/com/faforever/api/clan/ClansController.java
+++ b/src/main/java/com/faforever/api/clan/ClansController.java
@@ -1,5 +1,8 @@
 package com.faforever.api.clan;
 
+import com.faforever.api.clan.result.ClanResult;
+import com.faforever.api.clan.result.MeResult;
+import com.faforever.api.clan.result.PlayerResult;
 import com.faforever.api.data.domain.Clan;
 import com.faforever.api.data.domain.Player;
 import com.faforever.api.player.PlayerService;
@@ -40,24 +43,25 @@ public class ClansController {
       @ApiResponse(code = 200, message = "Success with JSON { player: {id: ?, login: ?}, clan: { id: ?, name: ?, tag: ?}}"),
       @ApiResponse(code = 400, message = "Bad Request")})
   @RequestMapping(path = "/me", method = RequestMethod.GET, produces = MediaType.APPLICATION_JSON_VALUE)
-  public Map<String, Serializable> me(Authentication authentication) {
+  public MeResult me(Authentication authentication) {
     Player player = playerService.getPlayer(authentication);
 
-    ImmutableMap<String, Serializable> playerMap = ImmutableMap.of(
-        "id", player.getId(),
-        "login", player.getLogin());
+    MeResult result = new MeResult();
+
+    result.setPlayer(new PlayerResult()
+        .setId(player.getId())
+        .setLogin(player.getLogin()));
 
     Clan clan = (!player.getClanMemberships().isEmpty())
         ? player.getClanMemberships().get(0).getClan()
         : null;
-    Serializable clanData = "";
     if (clan != null) {
-      clanData = ImmutableMap.of(
-          "id", clan.getId(),
-          "name", clan.getName(),
-          "tag", clan.getTag());
+      result.setClan(new ClanResult()
+          .setId(clan.getId())
+          .setName(clan.getName())
+          .setTag(clan.getTag()));
     }
-    return ImmutableMap.of("player", playerMap, "clan", clanData);
+    return result;
   }
 
   // This request cannot be handled by JSON API because we must simultaneously create two resources (a,b)
--- a/src/main/java/com/faforever/api/clan/ClansController.java
+++ b/src/main/java/com/faforever/api/clan/ClansController.java
@@ -46,22 +46,14 @@ public class ClansController {
   public MeResult me(Authentication authentication) {
     Player player = playerService.getPlayer(authentication);
 
-    MeResult result = new MeResult();
-
-    result.setPlayer(new PlayerResult()
-        .setId(player.getId())
-        .setLogin(player.getLogin()));
-
     Clan clan = (!player.getClanMemberships().isEmpty())
         ? player.getClanMemberships().get(0).getClan()
         : null;
+    ClanResult clanResult = null;
     if (clan != null) {
-      result.setClan(new ClanResult()
-          .setId(clan.getId())
-          .setName(clan.getName())
-          .setTag(clan.getTag()));
+      clanResult = new ClanResult(clan.getId(), clan.getTag(), clan.getName());
     }
-    return result;
+    return new MeResult(new PlayerResult(player.getId(), player.getLogin()), clanResult);
   }
 
   // This request cannot be handled by JSON API because we must simultaneously create two resources (a,b)
--- a/src/main/java/com/faforever/api/clan/result/ClanResult.java
+++ b/src/main/java/com/faforever/api/clan/result/ClanResult.java
@@ -1,0 +1,10 @@
+package com.faforever.api.clan.result;
+
+import lombok.Data;
+
+@Data
+public class ClanResult {
+  int id;
+  String name;
+  String tag;
+}
--- a/src/main/java/com/faforever/api/clan/result/ClanResult.java
+++ b/src/main/java/com/faforever/api/clan/result/ClanResult.java
@@ -4,7 +4,7 @@ import lombok.Data;
 
 @Data
 public class ClanResult {
-  int id;
-  String name;
-  String tag;
+  private final int id;
+  private final String tag;
+  private final String name;
 }
--- a/src/main/java/com/faforever/api/clan/result/MeResult.java
+++ b/src/main/java/com/faforever/api/clan/result/MeResult.java
@@ -1,0 +1,12 @@
+package com.faforever.api.clan.result;
+
+import lombok.Data;
+
+@Data
+public class MeResult {
+  private PlayerResult player;
+  private ClanResult clan;
+
+
+}
+
--- a/src/main/java/com/faforever/api/clan/result/MeResult.java
+++ b/src/main/java/com/faforever/api/clan/result/MeResult.java
@@ -4,9 +4,7 @@ import lombok.Data;
 
 @Data
 public class MeResult {
-  private PlayerResult player;
-  private ClanResult clan;
-
-
+  private final PlayerResult player;
+  private final ClanResult clan;
 }
 
--- a/src/main/java/com/faforever/api/clan/result/PlayerResult.java
+++ b/src/main/java/com/faforever/api/clan/result/PlayerResult.java
@@ -4,6 +4,6 @@ import lombok.Data;
 
 @Data
 public class PlayerResult {
-  int id;
-  String login;
+  private final int id;
+  private final String login;
 }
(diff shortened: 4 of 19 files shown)


src/main/java/com/faforever/api/clan/result/ClanResult.java
```diff
@@ -0,0 +1,10 @@
+package com.faforever.api.clan.result;
+
+import lombok.Data;
+
+@Data
+public class ClanResult {
+  int id;
```
micheljung: private final

src/main/java/com/faforever/api/clan/result/MeResult.java
```diff
@@ -0,0 +1,12 @@
...
+
+import lombok.Data;
+
+@Data
+public class MeResult {
+  private PlayerResult player;
+  private ClanResult clan;
+
+
+}
```
micheljung: private final, and remove empty lines

src/main/java/com/faforever/api/clan/result/PlayerResult.java
```diff
@@ -0,0 +1,9 @@
+package com.faforever.api.clan.result;
+
+import lombok.Data;
+
+@Data
+public class PlayerResult {
+  int id;
+  String login;
```
micheljung: private final

src/main/java/com/faforever/api/config/GlobalControllerExceptionHandler.java
```diff
@@ -0,0 +1,55 @@
...
+  @ExceptionHandler(ValidationException.class)
+  @ResponseStatus(HttpStatus.UNPROCESSABLE_ENTITY)
+  @ResponseBody
+  public Map<String, Serializable> processValidationException(ValidationException ex) {
+    return errorResponse(ErrorCode.VALIDATION_FAILED.getTitle(), ex.getMessage());
+  }
+
+  private String replaceArgs(String message, Object[] args) {
+    for (int i = 0; i < args.length; i++) {
+      message = message.replace("{" + i + "}", args[i].toString());
```
micheljung: * If `args[i]` is `null` (for whatever reason) this throws an NPE, which is why the null-safe `String.valueOf(Object)` should be used instead.
* However, `{0}` and such is part of the Java MessageFormat, which is why `MessageFormat.format(String, Object...)` should be used. So this method should be removed (MessageFormat does much more than just replace {n}, too)

src/main/java/com/faforever/api/config/GlobalControllerExceptionHandler.java
```diff
@@ -0,0 +1,55 @@
...
+  @ExceptionHandler(ApiException.class)
+  @ResponseStatus(HttpStatus.UNPROCESSABLE_ENTITY)
+  @ResponseBody
+  public ErrorResponse processApiException(ApiException ex) {
+    ErrorResponse response = new ErrorResponse();
+    Arrays.stream(ex.getErrors()).forEach(error -> {
+      ErrorCode code = error.getErrorCode();
+      response.addError(new ErrorResult()
+          .setTitle(code.getTitle())
+          .setDetail(replaceArgs(code.getDetail(), error.getArgs())));
```
micheljung: Use `Message.format()`

src/main/java/com/faforever/api/config/error/ErrorResponse.java
```diff
@@ -0,0 +1,15 @@
+package com.faforever.api.config.error;
+
+import lombok.Data;
+
+import java.util.ArrayList;
+
+@Data
+public class ErrorResponse {
+  private ArrayList<ErrorResult> errors = new ArrayList<>();
```
micheljung: Make fields that never change `final` (I feel like I'm repeating myself in every review)

src/main/java/com/faforever/api/config/error/ErrorResult.java
```diff
@@ -0,0 +1,9 @@
+package com.faforever.api.config.error;
+
+import lombok.Data;
+
+@Data
+public class ErrorResult {
+  private String title;
+  private String detail;
```
micheljung: `final`

src/test/java/com/faforever/api/clan/ClanControllerTest.java
```diff
@@ -0,0 +1,284 @@
...
+  private ClanMembershipRepository clanMembershipRepository;
+  private PlayerRepository playerRepository;
+  private OAuthClientRepository oAuthClientRepository;
+  private ObjectMapper objectMapper;
+  private ShaPasswordEncoder shaPasswordEncoder;
+  private Player me;
+
+  public ClanControllerTest() {
+    objectMapper = new ObjectMapper();
+    shaPasswordEncoder = new ShaPasswordEncoder(256);
```
micheljung: Constructor initialized fields that never change -> `final`

src/test/java/com/faforever/api/clan/ClanControllerTest.java
```diff
@@ -0,0 +1,284 @@
...
+    this.clanMembershipRepository = clanMembershipRepository;
+  }
+
+  @Before
+  public void setUp() {
+    mvc = MockMvcBuilders
+        .webAppContextSetup(context)
+        .addFilter(springSecurityFilterChain)
+        .build();
+    me = null;
```
micheljung: A new instance of the test class is instantiated before every `@Test`, which means `me` is always `null` when `setUp()` is called, which means this line is redundant.

src/test/java/com/faforever/api/clan/ClanControllerTest.java
```diff
@@ -0,0 +1,284 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.content;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest
+public class ClanControllerTest {
```
micheljung: Please suffix with `IntegrationTest`

src/main/java/com/faforever/api/clan/ClansController.java
```diff
@@ -40,24 +43,25 @@ public ClansController(ClanService clanService, PlayerService playerService) {
...
     Player player = playerService.getPlayer(authentication);
 
-    ImmutableMap<String, Serializable> playerMap = ImmutableMap.of(
-        "id", player.getId(),
-        "login", player.getLogin());
+    MeResult result = new MeResult();
+
+    result.setPlayer(new PlayerResult()
+        .setId(player.getId())
+        .setLogin(player.getLogin()));
```
micheljung: In order to ensure that clients can use the same mechanism to deserialize any object, we need to stick to the JSON-API.

You can achieve this like so:
```
    return new JsonApiDocument(new Data<>(new Resource("player", String.valueOf(player.getId()),
        ImmutableMap.of(
            "login", player.getName()
        ), null, null, null)));
```
(and attach the clan below accordingly).

However, in this case I still think there is a way to directly return `/player/1234`. Otherwise the client may fail to deserialize because the format doesn't match, but the type is the same. If we can't do that, we need to use another type than `player`.

src/main/java/com/faforever/api/data/checks/IsClanMembershipDeletable.java
```diff
@@ -9,16 +9,18 @@
...
   public static class Inline extends InlineCheck<ClanMembership> {
 
     @Override
     public boolean ok(ClanMembership membership, RequestScope requestScope, Optional<ChangeSpec> changeSpec) {
       Object opaqueUser = requestScope.getUser().getOpaqueUser();
+      int requesterId = ((FafUserDetails) opaqueUser).getId();
       return opaqueUser instanceof FafUserDetails
-          && membership.getClan().getLeader().getId() == ((FafUserDetails) opaqueUser).getId()
-          && membership.getPlayer().getId() != membership.getClan().getLeader().getId();
+          && membership.getPlayer().getId() != membership.getClan().getLeader().getId()
```
micheljung: Don't use `==` for String comparison. Use `Objects.compare()`

src/test/java/com/faforever/api/clan/ClanControllerTest.java
IDragonfire: `@IntegrationTest` is deprecated
image

src/main/java/com/faforever/api/clan/ClansController.java
IDragonfire: This is a special route, not a json api route !

src/main/java/com/faforever/api/data/checks/IsClanMembershipDeletable.java
IDragonfire: This compare ints oO

src/main/java/com/faforever/api/clan/result/ClanResult.java
IDragonfire: done

src/main/java/com/faforever/api/clan/result/MeResult.java
IDragonfire: done

src/main/java/com/faforever/api/clan/result/PlayerResult.java
IDragonfire: done

src/main/java/com/faforever/api/config/error/ErrorResponse.java
IDragonfire: done

src/main/java/com/faforever/api/config/error/ErrorResult.java
IDragonfire: done

src/test/java/com/faforever/api/clan/ClanControllerTest.java
IDragonfire: done

src/test/java/com/faforever/api/clan/ClanControllerTest.java
IDragonfire: done

src/main/java/com/faforever/api/config/GlobalControllerExceptionHandler.java
IDragonfire: http://stackoverflow.com/a/10995800 
`MessageFormat.format`need special escaping ...

src/main/java/com/faforever/api/data/checks/IsClanMembershipDeletable.java
micheljung: Oops, my bad, except that IDs in JSONAPI should always be strings :D But I think we can keep them as ints internally, and Elide will convert them as required (like we have been doing it as it seems).

src/test/java/com/faforever/api/clan/ClanControllerTest.java
micheljung: _Suffix_, not _annotate_

PR #59, opened 2017-05-01
Fix #57 Customized login page

IDragonfire:

PR #61, opened 2017-05-02
[2] Beautify Oauth Approval Page

IDragonfire:
Based on https://github.com/FAForever/faf-java-api/pull/59
only beautify the current login page

### DIFF
--- a/src/main/java/com/faforever/api/security/OAuthController.java
+++ b/src/main/java/com/faforever/api/security/OAuthController.java
@@ -1,0 +1,26 @@
+package com.faforever.api.security;
+
+import com.faforever.api.client.OAuthClient;
+import com.faforever.api.client.OAuthClientRepository;
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.servlet.ModelAndView;
+
+@Controller
+@RequestMapping("/oauth")
+public class OAuthController {
+  private OAuthClientRepository oAuthClientRepository;
+
+  public OAuthController(OAuthClientRepository oAuthClientRepository) {
+    this.oAuthClientRepository = oAuthClientRepository;
+  }
+
+  @GetMapping("/confirm_access")
+  public ModelAndView confirmAccess(@RequestParam(value = "client_id") String clientId) {
+    OAuthClient client = oAuthClientRepository.findOne(clientId);
+    return new ModelAndView("oauth_confirm_access").addObject("client", client);
+  }
+
+}
--- a/src/main/java/com/faforever/api/security/OAuthController.java
+++ b/src/main/java/com/faforever/api/security/OAuthController.java
@@ -9,7 +9,7 @@ import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.servlet.ModelAndView;
 
 import java.util.Arrays;
-import java.util.HashSet;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -27,10 +27,14 @@ public class OAuthController {
     @RequestParam("client_id") String clientId,
     @RequestParam("scope") String scope) {
     OAuthClient client = oAuthClientRepository.findOne(clientId);
-    final Set<OAuthScope> defaultScopes = getScopesFromScopeString(client.getDefaultScope());
+    Set<OAuthScope> scopes;
     final Set<OAuthScope> requestScopes = getScopesFromScopeString(scope);
-    Set<OAuthScope> scopes = new HashSet<>(defaultScopes);
-    scopes.addAll(requestScopes);
+    if (requestScopes.isEmpty()) {
+      final Set<OAuthScope> defaultClientScopes = getScopesFromScopeString(client.getDefaultScope());
+      scopes = defaultClientScopes;
+    } else {
+      scopes = requestScopes;
+    }
 
     return new ModelAndView("oauth_confirm_access")
       .addObject("client", client)
@@ -41,6 +45,8 @@ public class OAuthController {
     return Arrays.stream(scopeString.split(" "))
       .filter(s -> !s.isEmpty())
       .map(OAuthScope::fromKey)
+      .filter(Optional::isPresent)
+      .map(Optional::get)
       .collect(Collectors.toSet());
   }
 }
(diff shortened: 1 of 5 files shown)


IDragonfire:
@micheljung this should be merged as soon as possible, because the current authorization in productive looks really ugly ...

image

micheljung:
Make it display the actual scopes and I will merge it.

IDragonfire:
Help is welcome by @micheljung  or @bukajsytlos

src/main/resources/templates/oauth_confirm_access.html
```diff
@@ -0,0 +1,154 @@
...
+    }
+  </style>
+</head>
+
+<body>
+<div class="allow-card">
+  <h2>Grant Access</h2>
+  <p>This app would like to:</p>
+  <ul>
+    <li th:each="sc : ${#strings.listSplit(scope, ' ')}" th:switch="${sc}">
```
IDragonfire: This works, e.g. for `http://localhost:8010/oauth/authorize?client_id=83891c0c-feab-42e1-9ca7-515f94f808ef&redirect_uri=http%3A%2F%2Flocalhost%3A8080&scope=public_profile&response_type=token&state=` but in general we should read from the database, or? I have no idea ...

IDragonfire:
@micheljung what is the actual scope? The default scope of the database or the scope get parameter?

micheljung:
The "scope" in the DB is the allowed scope, the "default_scope" is what's taken as default if nothing else is requested in the HTTP request

bukajsytlos:
I am confused what do you mean by "scope" in DB. In DB I can only see default_scopes.
I modified code to merge request scopes with those default scopes. For now created enum of scopes, but maybe they should be defined somewhere else.

src/main/java/com/faforever/api/security/OAuthController.java
```diff
@@ -0,0 +1,52 @@
...
+    if (requestScopes.isEmpty()) {
+      final Set<OAuthScope> defaultClientScopes = getScopesFromScopeString(client.getDefaultScope());
+      scopes = defaultClientScopes;
+    } else {
+      scopes = requestScopes;
+    }
+
+    return new ModelAndView("oauth_confirm_access")
+      .addObject("client", client)
+      .addObject("scopes", scopes);
```
IDragonfire: This is not handled by the OAuth libary we use?
image
http://projects.spring.io/spring-security-oauth/docs/oauth2.html

Because it is a parameter http://docs.spring.io/spring-security/oauth/apidocs/org/springframework/security/oauth2/provider/client/BaseClientDetails.html#BaseClientDetails(java.lang.String 
Maybe it is enough to set it correctly

src/main/java/com/faforever/api/security/OAuthScope.java
```diff
@@ -0,0 +1,47 @@
...
+
+  PUBLIC_PROFILE("public_profile", "Read your public player data"),
+  READ_ACHIEVEMENTS("read_achievements", "Read your achievements"),
+  WRITE_ACHIEVEMENTS("write_achievements", "Write your achievements"),
+  READ_EVENTS("read_events", "Read events"),
+  WRITE_EVENTS("write_events", "Write events"),
+  UPLOAD_MAP("upload_map", "Upload maps"),
+  UPLOAD_MOD("upload_mod", "Upload mods"),
+  WRITE_ACCOUNT_DATA("write_account_data", "Edit account data"),
+  EDIT_CLAN_DATA("edit_clan_data", "Edit clan data");
```
IDragonfire: 👍 
We should then use this enums in the `@PreAuthorize` annotations

src/main/java/com/faforever/api/security/OAuthScope.java
```diff
@@ -0,0 +1,47 @@
...
+  public static Optional<OAuthScope> fromKey(String key) {
+    return Optional.ofNullable(fromString.get(key));
+  }
+
+  public String getKey() {
+    return key;
+  }
+
+  public String getTitle() {
+    return title;
```
IDragonfire: This is a more extendable and from my view better approach ;) Good work!

src/main/java/com/faforever/api/security/OAuthController.java
bukajsytlos: Not sure what are you referring. I put set of scopes in model to be able to list them in template.

IDragonfire: The scope must be put in the jwt token somehow, so somewhere the `merged` scope must exists, Or if no scope is provied the default scope is used, but this is incosistent to database changes.

IDragonfire:
@bukajsytlos I tested the bed3772591e74f50b4e60e856de112c278be92fe and it worked like a shame ;) Good work !

src/main/java/com/faforever/api/security/OAuthScope.java
micheljung: That's the problem; you can't use properties of enums in annotations. We have to use String constants.

micheljung: Unfortunately, you can't use properties of enums in annotations. We have to use String constants.

I propose that we do it like this:
```
public enum OAuthScope {
  public static final String PUBLIC_PROFILE = "public_profile";
  ...

  _PUBLIC_PROFILE("public_profile", "Read your public player data"),
  ...
}
```

This way we still have the scope and title grouped as an enum, but can also use `OAuthScope.PUBLIC_PROFILE` in the annotation.

src/main/java/com/faforever/api/security/OAuthScope.java
```diff
@@ -0,0 +1,47 @@
...
+
+  static {
+    fromString = new HashMap<>();
+    for (OAuthScope oAuthScope : values()) {
+      fromString.put(oAuthScope.key, oAuthScope);
+    }
+  }
+
+  private String key;
+  private String title;
```
micheljung: Make `final`

src/main/java/com/faforever/api/security/OAuthController.java
```diff
@@ -0,0 +1,52 @@
...
+
+import java.util.Arrays;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+@Controller
+@RequestMapping("/oauth")
+public class OAuthController {
+  private OAuthClientRepository oAuthClientRepository;
```
micheljung: Please look at `WhitelabelApprovalEndpoint` on how to do it properly.

src/main/java/com/faforever/api/security/OAuthController.java
micheljung: It's all handled by Spring, there is no need to merge scopes, no need to access the oauth client repository manually, no need to manually put anything into the JWT :) (which btw. stands for "JSON Web Token" so JWT _token_ is redundant ;))

IDragonfire: To display the permission messages we must only get somehow the scope ... no idea how ...

PR #84, opened 2017-05-09
[5] Name Entitiy

IDragonfire:
[Added by Brutus5000:] Depending on FAForever/db#137

PS.: codacy is wrong, method is used over lombok builder ...

### DIFF
--- a/src/test/java/com/faforever/api/data/JsonApiNameIntegrationTest.java
+++ b/src/test/java/com/faforever/api/data/JsonApiNameIntegrationTest.java
@@ -1,0 +1,93 @@
+package com.faforever.api.data;
+
+import com.faforever.api.data.domain.NameRecord;
+import com.faforever.api.data.domain.Player;
+import com.faforever.integration.TestDatabase;
+import com.faforever.integration.factories.NameRecordFactory;
+import com.faforever.integration.factories.PlayerFactory;
+import com.faforever.integration.utils.MockMvcHelper;
+import lombok.SneakyThrows;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.context.annotation.Import;
+import org.springframework.test.context.junit4.SpringRunner;
+import org.springframework.test.web.servlet.MockMvc;
+import org.springframework.test.web.servlet.setup.MockMvcBuilders;
+import org.springframework.web.context.WebApplicationContext;
+
+import javax.inject.Inject;
+import javax.servlet.Filter;
+
+import static org.hamcrest.Matchers.hasSize;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertEquals;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.jsonPath;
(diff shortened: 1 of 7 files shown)


bukajsytlos:
To which issue is this related?

IDragonfire:
No Issue, general needed, e.g. for name change ...

src/main/java/com/faforever/api/data/domain/Name.java
```diff
@@ -0,0 +1,47 @@
...
+public class Name {
+  private OffsetDateTime changeTime;
+  private Player player;
+  private String name;
+
+  @Id
+  @Column(name = "change_time")
+  public OffsetDateTime getChangeTime() {
+    return changeTime;
+  }
```
micheljung: We need to refactor the DB to have an `int id` primary key.

src/main/java/com/faforever/api/data/domain/Name.java
```diff
@@ -0,0 +1,47 @@
...
+import javax.validation.constraints.NotNull;
+import java.time.OffsetDateTime;
+
+@Entity
+@Table(name = "name_history")
+@Include(rootLevel = true, type = "name")
+@DeletePermission(expression = "Prefab.Role.None")
+@UpdatePermission(expression = "Prefab.Role.None")
+@Setter
+public class Name {
```
micheljung: An entity `Name` with an attribute `name` is rather unintuitive :)
How about any of `NameChange`, `PreviousName`, `HistoricName`, `NameRecord`, `TrackedName`, `PreviousName`?

src/main/java/com/faforever/api/data/domain/Name.java
IDragonfire: Issue created https://github.com/FAForever/db/issues/119

src/main/java/com/faforever/api/data/domain/Name.java
IDragonfire: I like `NameRecord` maybe `NicknameRecord` or `Nickname`?

micheljung: `NameRecord` sounds good. We never use the term `Nickname` anywhere, so we shouldn't introduce that. **If** anything, then it would be `UsernameRecord`.

src/main/java/com/faforever/api/data/domain/Name.java
Brutus5000: Seriously @IDragonfire it would haven taken you 5 more minutes to make the db and this would have been in production for a long time... https://github.com/FAForever/db/pull/137

IDragonfire: If I remeber correctly this is not the only place ...

Brutus5000:
I rebased this PR, referenced the new issue and applied all proposed changes. This is ready to merge, as soon as the corresponding db-PR is done.

src/test/java/com/faforever/api/data/JsonApiNameIntegrationTest.java
```diff
@@ -0,0 +1,93 @@
...
+  public void tearDown() {
+    // TODO: This is needed, because Elide has some problems with @Transactional annotation #71
+    database.tearDown();
+  }
+
+
+  @Test
+  @SneakyThrows
+  public void getNames() {
+    // TODO: use new PlayerFactor
```
micheljung: Obsolet?

PR #70, opened 2017-05-03
Fix #66 Don't manually read JSON values, deserialize into a Java object

IDragonfire:
Fix #66 Don't manually read JSON values, deserialize into a Java object

PR #85, opened 2017-05-10
Feature/#35 password change

Geosearchef:


src/main/java/com/faforever/api/user/UserService.java
```diff
@@ -124,4 +129,75 @@ void activate(String token) {
...
     userRepository.save(user);
   }
+
+  void changePassword(String newPassword, User user) {
+    user.setPassword(passwordEncoder.encode(newPassword));
+    userRepository.save(user);
+  }
+
+  void changeLogin(String newLogin, User user) {
+    user.setLogin(newLogin);
```
IDragonfire: Create an entry in `name_history` table (https://github.com/FAForever/faf-java-api/pull/84)
Optional: You cannot use an name that was used 6 months ago: https://github.com/FAForever/api/pull/163

src/main/java/com/faforever/api/user/UserController.java
```diff
@@ -26,8 +28,34 @@ public void register(@RequestParam("username") String username,
...
+  @PreAuthorize("#oauth2.hasScope('change_login') and hasRole('ROLE_USER')")
+  @ApiOperation("Changes the login of a previously registered account.")
+  @RequestMapping(path = "/changeLogin", method = RequestMethod.POST, produces = MediaType.APPLICATION_JSON_VALUE)
+  public void changeLogin(@RequestParam("newLogin") String newLogin, Authentication authentication) {
+    userService.changeLogin(newLogin, userService.getUser(authentication));
+  }
+
+  @ApiOperation("Sends a password reset to the email linked by this account.")
+  @RequestMapping(path = "/resetPassword", method = RequestMethod.POST, produces = MediaType.APPLICATION_JSON_VALUE)
+  public void resetPassword(@RequestParam("email") String email) {
```
IDragonfire: The old API requires correct `login` and `email` https://github.com/FAForever/api/blob/develop/api/users_route.py#L141

src/main/java/com/faforever/api/email/EmailService.java
```diff
@@ -50,4 +51,16 @@ public void sendActivationMail(String username, String email, String activationU
         String.format(registration.getHtmlFormat(), username, activationUrl)
     );
   }
+
+  @SneakyThrows
+  public void sendPasswordResetMail(String username, String email, String passwordResetUrl) {
+    PasswordReset passwordReset = properties.getPasswordReset();
+    emailSender.sendMail(
```
IDragonfire: PS.: I discovered latly the `@Builder` lombok annotation: https://projectlombok.org/features/Builder.html

src/main/java/com/faforever/api/user/UserService.java
```diff
@@ -124,4 +129,75 @@ void activate(String token) {
...
+
+  @SneakyThrows
+  void claimPasswordResetToken(String token, String newPassword) {
+    HashMap<String, String> claims = objectMapper.readValue(JwtHelper.decodeAndVerify(token, macSigner).getClaims(), HashMap.class);
+
+    String action = claims.get(KEY_ACTION);
+    if (!Objects.equals(action, ACTION_RESET_PASSWORD)) {
+      throw new ApiException(new Error(ErrorCode.TOKEN_INVALID));
+    }
+    if (Instant.parse(claims.get(KEY_EXPIRY)).isBefore(Instant.now())) {
```
IDragonfire: This is better then my System comparision ^^
Good to know
```java
@Data
public class InvitationResult {
  private final long expire;
  private final ClanResult clan;
  private final PlayerResult newMember;

  public boolean isExpired() {
    return expire < System.currentTimeMillis();
  }
}
```

src/main/java/com/faforever/api/user/UserService.java
```diff
@@ -124,4 +129,75 @@ void activate(String token) {
...
+
+    if (user == null) {
+      throw new ApiException(new Error(ErrorCode.TOKEN_INVALID));
+    }
+
+    user.setPassword(passwordEncoder.encode(newPassword));
+    userRepository.save(user);
+  }
+
+  public User getUser(Authentication authentication) {
```
IDragonfire: This should be moved in the `UserService`, similiar to https://github.com/FAForever/faf-java-api/blob/develop/src/main/java/com/faforever/api/player/PlayerService.java#L20

src/test/java/com/faforever/api/user/UserServiceTest.java
```diff
@@ -162,4 +164,118 @@ public void activateTokenExpired() throws Exception {
...
+        UserService.KEY_ACTION, "reset_password",
+        UserService.KEY_EXPIRY, Instant.now().plusSeconds(-100).toString(),
+        UserService.KEY_USER_ID, "5"
+    )), new MacSigner("banana")).getEncoded();
+
+    instance.claimPasswordResetToken(token, "newPassword");
+  }
+
+
+  private User dummyUser() {
```
IDragonfire: Consider creating a `Factory/Builder` for it ...https://github.com/FAForever/faf-java-api/tree/develop/src/test/java/com/faforever/integration/factories

src/main/java/com/faforever/api/user/UserService.java
```diff
@@ -25,23 +27,25 @@
 @Slf4j
 public class UserService {
 
-  private static final Pattern USERNAME_PATTERN = Pattern.compile("[A-Za-z][A-Za-z0-9_-]{2,15}$");
   static final String KEY_ACTION = "action";
   static final String KEY_EXPIRY = "expiry";
   static final String KEY_USERNAME = "username";
   static final String KEY_EMAIL = "email";
   static final String KEY_PASSWORD = "password";
+  static final String KEY_USER_ID = "id";
```
Geosearchef: This case is no dynamic data, it's constants for the keys of the token. I don't see why this should be in another class and clunk up space.

src/main/java/com/faforever/api/config/FafApiProperties.java
```diff
@@ -160,4 +161,14 @@
     private String fromEmail;
     private String fromName;
   }
+
+  @Data
+  public static class PasswordReset {
+    private int linkExpirationSeconds = 3600 * 24;
```
Geosearchef: This is the same as activation email data. I will change that aswell.

IDragonfire: +1

src/main/java/com/faforever/api/user/UserService.java
IDragonfire: What you mean with dynamic? 
Why is another class a problem? 
It should reduce the code and make it more readable
https://github.com/FAForever/faf-java-api/pull/70

src/main/java/com/faforever/api/user/UserController.java
```diff
@@ -26,8 +28,34 @@ public void register(@RequestParam("username") String username,
...
+  @RequestMapping(path = "/activate", method = RequestMethod.GET, produces = MediaType.APPLICATION_JSON_VALUE)
   public void activate(@RequestParam("token") String token) {
     userService.activate(token);
   }
+
+  @PreAuthorize("#oauth2.hasScope('change_password') and hasRole('ROLE_USER')")
+  @ApiOperation("Changes the password of a previously registered account.")
+  @RequestMapping(path = "/changePassword", method = RequestMethod.POST, produces = MediaType.APPLICATION_JSON_VALUE)
+  public void changePassword(@RequestParam("newPassword") String newPassword, Authentication authentication) {
+    userService.changePassword(newPassword, userService.getUser(authentication));
```
Brutus5000: Yes, but it should require the original password as a parameter. Oauth token alone is not safe .

src/main/java/com/faforever/api/user/UserController.java
Brutus5000: Yes, this is like an additional security question.

src/main/java/com/faforever/api/user/UserService.java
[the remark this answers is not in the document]
Geosearchef: This get's a bit weird in this case, as there may be more tokens signed by the server that contain only a userID, therefore the token needs to contain a String identifying the action. This gets a bit weird in object orientation, as a class now contains a String the is normally unnecessary.

Brutus5000:
Due to the lack of progress I am now officialy hijacking this PR.

src/main/java/com/faforever/api/user/UserController.java
Brutus5000: According to the many feedback we received in the last month it should be either `login` or `email`. It seems to be incredibly difficult for people to remember their username and/or email address...

Brutus5000:
I have added the following features on top of geosearchefs code (that helped a lot!):
- Password reset now takes a username OR an email
- NameRecords are saved on username change
- Username change restriction for 30 days
- Old usernames are reserved for 6 months (blocked in registration and username changes of other users)

I suggest a new review. There is probably still room for quality improvement.

src/main/java/com/faforever/api/user/UserService.java
```diff
@@ -124,4 +140,104 @@ void activate(String token) {
...
+    int minDaysBetweenChange = properties.getUser().getMinimumDaysBetweenUsernameChange();
+    nameRecordRepository.getDaysSinceLastNewRecord(user.getId(), minDaysBetweenChange)
+      .ifPresent(daysSinceLastRecord -> {
+        throw new ApiException(new Error(ErrorCode.USERNAME_CHANGE_TOO_EARLY, minDaysBetweenChange - daysSinceLastRecord));
+      });
+
+    int usernameReservationTimeInMonths = properties.getUser().getUsernameReservationTimeInMonths();
+    nameRecordRepository.getLastUsernameOwnerWithinMonths(newLogin, usernameReservationTimeInMonths)
+      .ifPresent(reservedByUserId -> {
+        if (reservedByUserId != user.getId()) {
```
micheljung: This is an object comparison that needs to be done using `equals()`

Brutus5000:
**Additional requirement from the #production-server slack**
Change the password in the anope db too, like the server currently does on login: 
https://github.com/FAForever/server/blob/dc963f4e0568bc3703012b5c28ed78422170c34c/server/lobbyconnection.py#L632

src/main/java/com/faforever/api/user/UserService.java
Brutus5000: Why? We are comparing to ints.

src/main/java/com/faforever/api/error/ErrorCode.java
```diff
@@ -67,7 +67,9 @@
   VALIDATION_FAILED(158, "Validation failed", "{0}"),
   MOD_UID_EXISTS(159, "Duplicate mod UID", "A mod with UID ''{0}'' already exists."),
   MOD_STRUCTURE_INVALID(160, "Invalid file structure for mod", "Files in the the root level of the zip file are not allowed. Please ensure all files reside inside a folder."),
-  MOD_VERSION_NOT_A_NUMBER(161, "Mod version is not a number", "The mod version has to be a whole number like 123, but was ''{0}''");
+  MOD_VERSION_NOT_A_NUMBER(161, "Mod version is not a number", "The mod version has to be a whole number like 123, but was ''{0}''"),
+  USERNAME_RESERVED(162, "Invalid account data", "The entered username is currently reserved: {0} (Maximum reservation time is {1} months)"),
```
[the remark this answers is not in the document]
Brutus5000: I will use your proposal as a text.
`Invalid account data` was introduced by yourself for EMAIL_INVALID and USERNAME_INVALID, .... This is just conformity.
If you make a suggestion I can change all of them.

src/main/java/com/faforever/api/user/UserService.java
micheljung: No, we are comparing two `Integer`. `==` will thus compare references.

Brutus5000: `reservedByUserId` is an Integer, while user.getId() returns `int`
If I  use `reservedByUserId.intValue()` IntelliJ starts complaining, so I leave it as is.

PR #89, opened 2017-05-13
[1] Cleanup Travis Log

IDragonfire:
* Colored test output
* Fold test output
* ~Reduce log~

With color and folding: https://travis-ci.org/FAForever/faf-java-api/builds/231914410
Without `--info`: https://travis-ci.org/FAForever/faf-java-api/builds/231910008

PR #92, opened 2017-05-14
Use client extension for mod package files

duk3luk3:
Makes sure deployed mod package files do not conflict.

This is necessary for the legacy patcher, since it creates patch files using the scheme <old-filename>-<new-filename> and overwrites old files, and then delivers patch files that cannot be applied.

There are two other ways of getting rid of the bug:

* Update the legacy patcher to stop creating conflicting patch files
* Stop using the legacy patcher altogether

The latter option is the most preferable but cannot be done in the short term. The former action means updating a component that is slated to be eliminated in the medium term and that nobody understands.

Therefore, making this simple, low-risk change to the api is the best choice.


Sheeo:
@micheljung got an idea of how to make CI output the test failure details in the log, rather than generating an HTML report?

IDragonfire:
I submited an PR already 2 days ago
https://github.com/FAForever/faf-java-api/pull/89

duk3luk3:
That's what I expected - thanks!

micheljung:
Woops, sorry for merging prematurely

Issue #112, opened 2017-06-20
Properly handle "user not found"


micheljung:
```
Caused by: java.lang.NullPointerException: null
	at com.faforever.api.security.FafUserDetails.<init>(FafUserDetails.java:19) ~[faf-java-api-0.5.9.1.jar:na]
	at com.faforever.api.security.FafUserDetailsService.loadUserByUsername(FafUserDetailsService.java:26) ~[faf-java-api-0.5.9.1.jar:na]
	at org.springframework.security.authentication.dao.DaoAuthenticationProvider.retrieveUser(DaoAuthenticationProvider.java:114) ~[spring-security-core-4.2.2.RELEASE.jar:4.2.2.RELEASE]
	... 51 common frames omitted
```

Because the user could not be found:
```
    return new FafUserDetails(userRepository.findOneByLoginIgnoreCase(username));
```
Should throw a `UsernameNotFoundException` instead

Yogendra0Sharma:
@micheljung Hi, I think the method loadUserByUsername()  in FafUserDetailsService.java already throws UsernameNotFoundException.

micheljung:
@Yogendra0Sharma it only declares it, but it needs to actually throw it if `findOneByLoginIgnoreCase()` returns null

ahsanbagwan:
Looks like the code coverage suggests adding some tests. Is that so? Please suggest.

PR #124, opened 2017-08-03
Properly handle "user not found" #112

ahsanbagwan:
Added the conditional including throwing the relevant exception.

Unless there is anything else that needs to be done, please merge :)

### DIFF
--- a/src/main/java/com/faforever/api/security/FafUserDetailsService.java
+++ b/src/main/java/com/faforever/api/security/FafUserDetailsService.java
@@ -23,9 +23,10 @@ public class FafUserDetailsService implements UserDetailsService {
 
   @Override
   public UserDetails loadUserByUsername(String username) throws UsernameNotFoundException {
-    if (userRepository.findOneByLoginIgnoreCase(username) == null) {
-      throw new UsernameNotFoundException("Username can't be null");
+    String uname = userRepository.findOneByLoginIgnoreCase(username); 
+    if (uname == null) {
+      throw new UsernameNotFoundException("User could not be found: "+username);
     }
-    return new FafUserDetails(userRepository.findOneByLoginIgnoreCase(username));
+    return new FafUserDetails(userRepository.findOneByLoginIgnoreCase(uname));
   }
 }
--- a/src/main/java/com/faforever/api/security/FafUserDetailsService.java
+++ b/src/main/java/com/faforever/api/security/FafUserDetailsService.java
@@ -1,5 +1,6 @@
 package com.faforever.api.security;
 
+import com.faforever.api.data.domain.User;
 import com.faforever.api.user.UserRepository;
 import org.springframework.security.core.userdetails.UserDetails;
 import org.springframework.security.core.userdetails.UserDetailsService;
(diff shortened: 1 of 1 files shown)


src/main/java/com/faforever/api/security/FafUserDetailsService.java
```diff
@@ -22,7 +22,10 @@ public FafUserDetailsService(UserRepository userRepository) {
   }
 
   @Override
-  public UserDetails loadUserByUsername(String username) throws UsernameNotFoundException {
+  public UserDetails loadUserByUsername(String username) {
```
Brutus5000: `throws UsernameNotFoundException` should not be removed

src/main/java/com/faforever/api/security/FafUserDetailsService.java
```diff
@@ -23,6 +23,9 @@ public FafUserDetailsService(UserRepository userRepository) {
 
   @Override
   public UserDetails loadUserByUsername(String username) throws UsernameNotFoundException {
+    if (userRepository.findOneByLoginIgnoreCase(username) == null) {
```
micheljung: `userRepository.findOneByLoginIgnoreCase(username)` is called twice within this method, which potentially causes two database calls.

Assign the result to a variable instead (Select statement -> CTRL+ALT+V) and use the variable instead. This way there is only one databse call.

src/main/java/com/faforever/api/security/FafUserDetailsService.java
```diff
@@ -23,6 +23,9 @@ public FafUserDetailsService(UserRepository userRepository) {
...
+      throw new UsernameNotFoundException("Username can't be null");
```
micheljung: This error message is misleading; it's not that the username is null, but that the username could not be found.
`"User could not be found: " + username` would be a proper message.

src/main/java/com/faforever/api/security/FafUserDetailsService.java
```diff
@@ -23,6 +24,10 @@ public FafUserDetailsService(UserRepository userRepository) {
 
   @Override
   public UserDetails loadUserByUsername(String username) throws UsernameNotFoundException {
-    return new FafUserDetails(userRepository.findOneByLoginIgnoreCase(username));
+    User uname = userRepository.findOneByLoginIgnoreCase(username);
```
micheljung: Name the variable `user` instead of `uname`, since it contains a user and not a user name.

src/main/java/com/faforever/api/security/FafUserDetailsService.java
```diff
@@ -23,6 +24,10 @@ public FafUserDetailsService(UserRepository userRepository) {
...
   public UserDetails loadUserByUsername(String username) throws UsernameNotFoundException {
-    return new FafUserDetails(userRepository.findOneByLoginIgnoreCase(username));
+    User uname = userRepository.findOneByLoginIgnoreCase(username); 
+    if (uname == null) {
+      throw new UsernameNotFoundException("User could not be found: "+username);
+    }
+    return new FafUserDetails(uname);
   }
-}
+}
```
micheljung: In IntelliJ's commit dialog, please enable the following settings under "Before Commit":
* Reformat code
* Rearrange code
* Optimize imports

This will ensure that the code is properly formatted (which, currently, it's not :))

PR #139, opened 2017-09-27
Feature/#81 legacy permission

Brutus5000:
This PR adds the permission system based on the existing `lobby_admin` table groups.
As a first reference I have prepared the `ladder1v1Map` entity for changing by faf moderators.
It uses an enhanced audit logger. This could be easily extended to write to file system, but the parameters are unclear to me.

Example logging:
```
c.f.api.security.ExtendedAuditLogger     : Removed map `SCMP_001` with version `1` from the ladder pool [invoked by user `Dostya` with id `2`]
c.f.api.security.ExtendedAuditLogger     : Added map `SCMP_001` with version `1` to the ladder pool [invoked by user `Dostya` with id `2`]
```

Any advice on how to add tests to this are welcome.

### DIFF
--- a/src/main/java/com/faforever/api/data/domain/LegacyAccessLevel.java
+++ b/src/main/java/com/faforever/api/data/domain/LegacyAccessLevel.java
@@ -4,6 +4,11 @@ import lombok.AllArgsConstructor;
 import lombok.Getter;
 import org.springframework.security.core.GrantedAuthority;
 
+import java.util.stream.Stream;
+
+/**
+ * @deprecated AccessLevel are going to be replaced with role based security
+ */
 @Getter
 @AllArgsConstructor
 @Deprecated
@@ -15,13 +20,10 @@ public enum LegacyAccessLevel implements GrantedAuthority {
   private final int code;
 
   public static LegacyAccessLevel fromCode(int code) {
-    for (LegacyAccessLevel level : LegacyAccessLevel.values()) {
-      if (level.code == code) {
-        return level;
-      }
-    }
-
-    throw new IllegalArgumentException(String.format("Code '%s' is unknown", code));
+    return Stream.of(LegacyAccessLevel.values())
+      .filter(level -> level.code == code)
+      .findFirst()
+      .orElseThrow(() -> new IllegalArgumentException(String.format("Code '%s' is unknown", code)));
   }
 
   @Override
--- a/src/main/java/com/faforever/api/data/domain/LegacyAccessLevel.java
+++ b/src/main/java/com/faforever/api/data/domain/LegacyAccessLevel.java
@@ -1,0 +1,31 @@
+package com.faforever.api.data.domain;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import org.springframework.security.core.GrantedAuthority;
+
+@Getter
+@AllArgsConstructor
+@Deprecated
+public enum LegacyAccessLevel implements GrantedAuthority {
+  ROLE_USER(0),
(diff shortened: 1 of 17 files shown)


src/main/java/com/faforever/api/data/domain/LobbyGroup.java
```diff
@@ -0,0 +1,41 @@
...
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.JoinColumn;
+import javax.persistence.OneToOne;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "lobby_admin")
+@Setter
+@Deprecated
```
micheljung: Explain why this is deprecated:
```
/**
 * @deprecated <reason>
 */
public class LobbyGroup
```

src/main/java/com/faforever/api/data/domain/LegacyAccessLevel.java
```diff
@@ -0,0 +1,31 @@
...
+
+@Getter
+@AllArgsConstructor
+@Deprecated
+public enum LegacyAccessLevel implements GrantedAuthority {
+  ROLE_USER(0),
+  ROLE_MODERATOR(1),
+  ROLE_ADMINISTRATOR(2);
+
+  private final int code;
```
micheljung: If we intend to always keep the code like an array index, we could dismiss this and just use the enum's  ordinal.
If the code could be an arbitrary number, then we should keep it.

src/main/java/com/faforever/api/data/domain/LegacyAccessLevel.java
```diff
@@ -0,0 +1,31 @@
...
+
+  public static LegacyAccessLevel fromCode(int code) {
+    for (LegacyAccessLevel level : LegacyAccessLevel.values()) {
+      if (level.code == code) {
+        return level;
+      }
+    }
+
+    throw new IllegalArgumentException(String.format("Code '%s' is unknown", code));
```
micheljung: Or short:
```
    return Stream.of(LegacyAccessLevel.values())
      .filter(level -> level.code == code)
      .findFirst()
      .orElseThrow(() -> new IllegalArgumentException(String.format("Code '%s' is unknown", code)));
```

src/main/java/com/faforever/api/security/FafUserDetailsService.java
```diff
@@ -25,8 +28,13 @@ public FafUserDetailsService(UserRepository userRepository) {
...
-      .orElseThrow(() -> new UsernameNotFoundException("User could not be found: " + username));
+    .orElseThrow(() -> new UsernameNotFoundException("User could not be found: " + username));
 
-    return new FafUserDetails(user);
+ArrayList<GrantedAuthority> authorities = new ArrayList<>();
+    authorities.add(LegacyAccessLevel.ROLE_USER);
+
+    if (user.getLobbyGroup() != null) {
+      authorities.add(LegacyAccessLevel.fromCode(user.getLobbyGroup().getAccessLevel()));
+    }    return new FafUserDetails(user, authorities);
```
micheljung: Formatting

src/main/java/com/faforever/api/data/domain/LegacyAccessLevel.java
Brutus5000: It's not intended as an array index. It just reflects the current values for these roles in the database (which are very unconvenient because you can't distinguish any further.

Brutus5000:
@micheljung Please check once more :)

PR #153, opened 2017-11-01
Migrate all integration test into separate module

Brutus5000:


build.gradle
```diff
@@ -147,11 +147,36 @@ processResources {
...
+  classpath = sourceSets.inttest.runtimeClasspath
+  mustRunAfter test
+}
+
 // JACOCO & COVERALLS
 
 apply plugin: 'jacoco'
 apply plugin: 'com.github.kt3k.coveralls'
 jacocoTestReport {
+  additionalSourceDirs = files(sourceSets.inttest.allJava.srcDirs)
```
micheljung: This treats the `inttest` directory as application source code to be covered. That's why the coverage dropped 20%.

src/main/java/com/faforever/api/data/domain/Message.java
```diff
@@ -24,7 +24,7 @@ public int getId() {
     return id;
   }
 
-  @Column(name = "key")
+  @Column(name = "`key`")
```
bukajsytlos: "\"key\"' should be db agnostic way of using keywords as column names

build.gradle
Brutus5000: No, the coverage dropped already before the last commit. This line was the bumbling try to raise it again.

PR #175, opened 2017-11-18
Closes #170 avatar management

### LINKED ISSUE #170: Avatar management
- [ ] Add a route for adding and uploading new avatars
- [ ] Make avatars deletable via Elide for moderators
      - check that it's not assigned before
      -  if valid, also delete file from disk
- [ ] Make avatar assignments creatable and deletable via Elide for moderators

bukajsytlos:


src/main/java/com/faforever/api/config/FafApiProperties.java
```diff
@@ -113,6 +115,15 @@
     private String downloadUrlFormat;
   }
 
+  @Data
+  public static class Avatar {
+    private String downloadUrlBase;
+    private List<String> allowedFileExtensions;
+    private Integer maxSizeBytes;
```
micheljung: `allowedFileExtensions` and `maxSizeBytes` should have default values (hint: Use `Arrays.asList()`)

src/main/java/com/faforever/api/data/domain/Avatar.java
```diff
@@ -25,12 +25,13 @@
   private List<AvatarAssignment> assignments;
 
   @Column(name = "url")
-  @NotNull
+//  @NotNull
```
micheljung: Why commented out? Add back, remove, or explain

src/main/java/com/faforever/api/data/domain/AvatarAssignment.java
```diff
@@ -25,26 +33,36 @@
...
   @JoinColumn(name = "idUser")
   @NotNull
+  @UpdatePermission(expression = "Prefab.Common.UpdateOnCreate")
   public Player getPlayer() {
     return player;
   }
+
+  @Override
+  @Transient
+  public Login getEntityOwner() {
```
micheljung: This should be annotated with `com.yahoo.elide.annotation.Exclude`. Best case we could specify in the interface, needs to be tested.

Otherwise, add it to all other implementations as well.

src/main/java/com/faforever/api/error/ErrorCode.java
```diff
@@ -71,7 +71,12 @@
   USERNAME_RESERVED(162, "Invalid account data", "The username ''{0}'' can only be claimed by the original owner within {1} months after it has been freed."),
   UNKNOWN_IDENTIFIER(163, "Unable to resolve user", "The identifier does neither match a username nor an email: {0}"),
   ALREADY_REGISTERED(164, "Registration failed", "You can't create a new account because you already have one."),
-  STEAM_LINK_NO_FA_GAME(141, "Linking to Steam failed", "You do not own Forged Alliance on Steam or your profile is private.");
+  STEAM_LINK_NO_FA_GAME(165, "Linking to Steam failed", "You do not own Forged Alliance on Steam or your profile is private."),
+  FILE_NAME_TOO_LONG(166, "Invalid file name", "The file name must not exceed {0} characters, was: {1}"),
+  FILE_SIZE_EXCEEDED(167, "Invalid file size", "The file size must not exceed {0} bytes, was: {1}"),
```
micheljung: The `{0}` should be `''{0}''` **however**, since it's a number, this is not needed **if** we can make sure that a number needs to be passed. For that, we can use `{0,number}` :)

Please also fix it at other places above, where I didn't do it.

src/main/java/com/faforever/api/error/ErrorCode.java
```diff
@@ -71,7 +71,12 @@
...
+  AVATAR_NAME_CONFLICT(168, "Invalid avatar file name", "Avatar file name ''{0}'' already exists."),
+  AVATAR_IN_USE(169, "Avatar in use", "Could not delete avatar. Avatar still in use."),
```
micheljung: `Could not delete avatar {0,number}`, with the avatar's ID.

src/main/java/com/faforever/api/map/MapsController.java
```diff
@@ -61,7 +61,7 @@ public void uploadMap(@RequestParam("file") MultipartFile file,
 
     String extension = Files.getFileExtension(file.getOriginalFilename());
     if (Arrays.stream(fafApiProperties.getMap().getAllowedExtensions()).noneMatch(extension::equals)) {
-      throw new ApiException(new Error(ErrorCode.UPLOAD_INVALID_FILE_EXTENSION, (Object[]) fafApiProperties.getMap().getAllowedExtensions()));
+      throw new ApiException(new Error(ErrorCode.UPLOAD_INVALID_FILE_EXTENSIONS, String.join(", ", fafApiProperties.getMap().getAllowedExtensions())));
```
micheljung: This way, we take the option away from the client to interpret the array of file extension differently. Keep it an array or explain why this is better :)

src/main/resources/config/application-dev.yml
```diff
@@ -14,6 +14,11 @@ faf-api:
     preview-url-format: ${MOD_PREVIEW_URL_FORMAT:http://test.content.faforever.com/faf/vault/mods/%s}
   replay:
     download-url-format: ${REPLAY_DOWNLOAD_URL_FORMAT:http://content.test.faforever.com/faf/vault/replay_vault/replay.php?id=%s}
+  avatar:
+    download-url-base: ${AVATAR_DOWNLOAD_URL_BASE:http://content.test.faforever.com/faf/avatars/}
+    allowed-file-extensions: ${AVATAR_ALLOWED_FILE_EXTENSIONS:png}
+    max-size-bytes: ${AVATAR_MAX_SIZE_BYTES:4096}
+    target-directory: ${AVATAR_TARGET_DIRECTORY:.}
```
micheljung: Please don't pollute our project directory :D Use `build/cache/avatars`

src/main/resources/config/application-prod.yml
```diff
@@ -15,6 +15,11 @@ faf-api:
     thumbnail-target-directory: ${MOD_THUMBNAIL_TARGET_DIRECTORY}
   replay:
     download-url-format: ${REPLAY_DOWNLOAD_URL_FORMAT}
+  avatar:
+    download-url-base: ${AVATAR_DOWNLOAD_URL_BASE}
+    allowed-file-extensions: ${AVATAR_ALLOWED_FILE_EXTENSIONS}
+    max-size-bytes: ${AVATAR_MAX_SIZE_BYTES}
```
micheljung: Let's define default values for `allowed-file-extensions` and `max-size-bytes`.

src/test/java/com/faforever/api/avatar/AvatarServiceTest.java
```diff
@@ -0,0 +1,151 @@
...
+public class AvatarServiceTest {
+
+  private static final long VALID_FILE_SIZE = 1024;
+  private static final long TOO_BIG_FILE_SIZE = 2048;
+  private static final String TOOLTIP = "Very nice avatar";
+  private static final String VALID_AVATAR_FILENAME = "qai2.png";
+  private static final String BIG_AVATAR_FILENAME = "donator.png";
+  private static final String LONG_AVATAR_FILENAME = "CachedAvataravatar_FightNight_Champion.png";
+  private static final String INVALID_EXTENSION_AVATAR_FILENAME = "supcom.jpg";
+  protected static final int AVATAR_ID = 1;
```
micheljung: No reason for `protected`?

src/main/java/com/faforever/api/avatar/AvatarService.java
```diff
@@ -0,0 +1,88 @@
...
+        log.warn("Could not delete file " + imageTargetPath, ioException);
+      }
+      throw e;
+    }
+  }
+
+  @SneakyThrows
+  @Transactional
+  public void deleteAvatar(Integer avatarId) {
+    final Avatar avatar = avatarRepository.getOne(avatarId);
```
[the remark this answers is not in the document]
bukajsytlos: so I should not focus on returning 404? Or should I "hack" exception handler that in case of this particular ApiException the return code should be 404?

src/main/java/com/faforever/api/avatar/AvatarService.java
```diff
@@ -0,0 +1,88 @@
...
+    final Avatar avatar = new Avatar()
+      .setTooltip(tooltip)
+      .setUrl(url);
+    try {
+      avatarRepository.save(avatar);
+    } catch (DataAccessException e) {
+      try {
+        Files.delete(imageTargetPath);
+      } catch (IOException ioException) {
+        log.warn("Could not delete file " + imageTargetPath, ioException);
```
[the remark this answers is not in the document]
bukajsytlos: I think it is not possible. there is no such a method

src/main/java/com/faforever/api/data/domain/Avatar.java
bukajsytlos: ah sry this is left over from some experiments

src/main/java/com/faforever/api/data/domain/AvatarAssignment.java
bukajsytlos: How should this manifest? I dont see it in elide responses?

src/test/java/com/faforever/api/avatar/AvatarServiceTest.java
bukajsytlos: yep sry for some reason refactor extract constant used protected

src/main/resources/config/application-prod.yml
bukajsytlos: just in Avatar properties class or here also?

src/main/java/com/faforever/api/map/MapsController.java
bukajsytlos: MessageFormat.format(ErrorCode.UPLOAD_INVALID_FILE_EXTENSIONS.getDetail(), (Object[]) new String[]{"png", "jpg"}) == "File must have the following extensions: png."

src/main/java/com/faforever/api/avatar/AvatarService.java
```diff
@@ -0,0 +1,88 @@
...
+    if (avatarImageFileSize > maxFileSizeBytes) {
+      throw new ApiException(new Error(ErrorCode.FILE_SIZE_EXCEEDED, maxFileSizeBytes, avatarImageFileSize));
+    }
+    if (properties.getAvatar().getAllowedFileExtensions().stream().noneMatch(extension::equals)) {
+      throw new ApiException(new Error(ErrorCode.UPLOAD_INVALID_FILE_EXTENSIONS, String.join(", ", properties.getAvatar().getAllowedFileExtensions())));
+    }
+    if (originalFilename.length() > maxFileNameLength) {
+      throw new ApiException(new Error(ErrorCode.FILE_NAME_TOO_LONG, maxFileNameLength, originalFilename.length()));
+    }
+  }
```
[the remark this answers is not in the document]
bukajsytlos: resolved in other part

src/main/java/com/faforever/api/avatar/AvatarService.java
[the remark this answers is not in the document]
Brutus5000: Yes you should. But actually you are not checking the avatar == null. And since you already declared `findById` and it returns an optional, it's easier to use the optional.

src/main/java/com/faforever/api/data/domain/AvatarAssignment.java
Brutus5000: Exactly. Should be like @JsonIgnore but for elide.

src/main/java/com/faforever/api/map/MapsController.java
Brutus5000: The cast to Object[] should not be necessary.

src/main/java/com/faforever/api/avatar/AvatarService.java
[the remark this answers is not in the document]
bukajsytlos: getOne will not return null. It will throw EntityNotFoundException which I handled as 404 in globalexceptionhandler and could be used by any other repository. In previous implementation I used findById but throw it away since throwing ApiException would result in 422 or I would need to hack handler method.
So what is the correct path I should choose?

src/main/java/com/faforever/api/data/domain/AvatarAssignment.java
bukajsytlos: I have no problem adding that but I am curious why since I cannot find entityOwner in AvatarAssigments response. Maybe @Transient is enough or I am blind :)

src/main/java/com/faforever/api/map/MapsController.java
bukajsytlos: ah now I get it :) I destroyed args.

src/main/java/com/faforever/api/data/domain/AvatarAssignment.java
bukajsytlos: https://github.com/yahoo/elide/blob/ef260b57ef9d8e6a403731dc79a5238bd3fa324c/elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java#L111

bukajsytlos: so either show me if you see owner in responses or I dont see much meaning in puting @Exclude anywhere

build.gradle
```diff
@@ -319,6 +319,7 @@ dependencies {
   compile("io.prometheus:simpleclient_servlet:${prometheusVersion}")
   compile("io.prometheus:simpleclient_spring_boot:${prometheusVersion}")
 
+  compile("com.github.jasminb:jsonapi-converter:${jsonapiConverterVersion}")
```
Brutus5000: I don't like to ship an end user library in the server. We are adding annotations into our domain classes just for the sake of integration testing. That adds unnecessary complexity, along with confusing overloading of @Id annotation.
We should go for static json strings instead.

@micheljung Your opinion on this?

src/inttest/java/com/faforever/api/avatar/AvatarControllerTest.java
```diff
@@ -0,0 +1,146 @@
...
+    Files.copy(FileHandlingHelper.loadResourceAsStream("/avatars/donator.png"), Paths.get("build/cache/avatars/avatar1.png"));
+    mockMvc.perform(
+      createAvatarReuploadRequest(1)
+        .with(getOAuthToken(OAuthScope._UPLOAD_AVATAR))
+    ).andExpect(status().isOk())
+      .andExpect(content().string(""));
+    final Avatar avatar = avatarRepository.findOneByUrl("http://localhost/faf/avatars/avatar1.png").get();
+    assertThat(avatar.getUrl(), is("http://localhost/faf/avatars/avatar1.png"));
+    assertThat(avatar.getTooltip(), is("Best avatar"));
+  }
```
Brutus5000: We should assert, that the file has actually changed on the file system.

src/inttest/java/com/faforever/api/avatar/AvatarControllerTest.java
```diff
@@ -0,0 +1,146 @@
...
+  @WithUserDetails(AUTH_MODERATOR)
+  public void moderatorCanUpload() throws Exception {
+    mockMvc.perform(
+      createAvatarUploadRequest()
+        .with(getOAuthToken(OAuthScope._UPLOAD_AVATAR))
+    ).andExpect(status().isCreated())
+      .andExpect(content().string(""));
+    final Avatar avatar = avatarRepository.findOneByUrl("http://localhost/faf/avatars/avatar3.png").get();
+    assertThat(avatar.getUrl(), is("http://localhost/faf/avatars/avatar3.png"));
+    assertThat(avatar.getTooltip(), is("Best avatar"));
```
Brutus5000: We should assert that the files exists on the file system.

src/inttest/java/com/faforever/api/avatar/AvatarControllerTest.java
```diff
@@ -0,0 +1,146 @@
...
+  @Test
+  @WithUserDetails(AUTH_MODERATOR)
+  public void moderatorCanDeleteAvatar() throws Exception {
+    Files.copy(FileHandlingHelper.loadResourceAsStream("/avatars/donator.png"), Paths.get("build/cache/avatars/avatar1.png"));
+    mockMvc.perform(
+      delete("/avatars/1")
+        .with(getOAuthToken(OAuthScope._UPLOAD_AVATAR))
+    ).andExpect(status().isNoContent());
+    assertThat(avatarRepository.findById(1), is(Optional.empty()));
+  }
```
Brutus5000: We should assert, that the file has been remove from file system.

bukajsytlos: this is tested in unit test of AvatarService

src/inttest/java/com/faforever/api/avatar/AvatarControllerTest.java
bukajsytlos: this is tested in unit test of AvatarService

src/inttest/java/com/faforever/api/avatar/AvatarControllerTest.java
bukajsytlos: this is tested in unit test of AvatarService

build.gradle
Brutus5000: Ok result from the discord discussion:
We want to use a separate repository (i.e. faf-java-commons) for the client-side dtos which have the jsonapi-converter annotations. These can be included with testcompile and used in the integration tests. These DTOs are needed in java-client and moderator-client anyway so testing them in the api gives additional test coverage.

PR #208, opened 2018-03-21
filename filter maps

1-alex98:


axel1200:
How do we go about it?

bukajsytlos:
Maybe you could use FileNameUtil. If you have windows environment you could run tests. @micheljung was complaining about test failing under windows. But it is working fine in linux env. Maybe regex can be tweaked to support both platforms' filenames

axel1200:
Where should I use FileNameUtil?

axel1200:
Can you comment on lines ?

axel1200:
The approach here is to just remove faulty characters here on upload

src/main/java/com/faforever/api/map/MapService.java
```diff
@@ -71,6 +71,8 @@ void uploadMap(byte[] mapData, String mapFilename, Player author, boolean isRank
     Assert.notNull(author, "'author' must not be null");
     Assert.isTrue(mapData.length > 0, "'mapData' must not be empty");
 
+    mapFilename = mapFilename.replaceAll("[^\\w.\\-]", ""); //replacing all characters that are illegal in filenames
```
bukajsytlos: consider using FileNameUtil.normalizeFileName method

axel1200:
I did

Brutus5000:
Due to current cases on prod we not only need to check the file name but also remove all non-ascii characters from the map name itself. @axel1200 Could you add this here?

axel1200:
Ehmm is that not what I was doing?

axel1200:
@bukajsytlos I looked into the normalize file name method and it is pretty bad it replaces %| and some other character instead of replacing everything that is not a letter or an underscore. because I could imagine some additional characters that wont get replaced like this and would cause problems again...

bukajsytlos:
@axel1200 yeah I take a look at it too. The thing is that it should produce valid filenames from OS stand-point of view. If this does not meet business requirement for map names we could create new name transformation. Not sure which characters should be removed for backward compatibility

axel1200:
tab for example should be removed

1-alex98:
for example today somebody uploaded map with : character in it and broke the vault

Brutus5000:
Short update: I am currently working on improvements of this PR

1-alex98:
Do so

Issue #140, opened 2017-10-02
Update to Elide 4.0 (when available)


micheljung:
They fixed https://github.com/yahoo/elide/issues/387

Brutus5000:
- [X]  SharePermission changed: remove the expression if it is not different from ReadPermission
- [X] HibernateStore removed, use AbstractHibernateStore and AbstractHibernateStore.Builder instead
- [ ] Fix error transaction now closed (maybe switch to Hibernate 5.2.x, but Spring Boot 1.5.x does not offer Hibernate 5.2.x)

PR #211, opened 2018-04-18
Updated Elide version

Rackover:
Elide throws errors on some record. As @micheljung suggested I updated the Elide version in the configuration file. Now we're gonna see if Travis can still build the thing.


Brutus5000:
Much more changes are required to make the api work with the new elide. Details see here https://github.com/FAForever/faf-java-api/issues/140
Since I guess you won't do it and the branch is in your fork, I close this PR.

PR #235, opened 2018-07-13
Bumpe Version to 1.7.0

1-alex98:


micheljung:
versions get "bumped" not "dumped" :P

axel1200:
I found some bug that should be fixed before release anyway

PR #237, opened 2018-07-21
Upgrade to Spring Boot 2 & Elide 4.2.3

Brutus5000:


src/main/java/com/faforever/api/config/security/WebSecurityConfig.java
```diff
@@ -45,9 +45,11 @@ public void developUserDetails(AuthenticationManagerBuilder auth) throws Excepti
 
   @Inject
   public void prodUserDetails(AuthenticationManagerBuilder auth, UserDetailsService userDetailsService) throws Exception {
+    // TODO: Use a proper, backward-compatible password encryption
```
micheljung: See https://github.com/FAForever/faf-java-server/blob/6709635794c43aacf3c8a1886766f28be102fe4a/faf-java-server-app/src/main/java/com/faforever/server/config/security/FafWebSecurityConfigurerAdapter.java#L35

Basically, you need to provide a mapping of which {PREFIX} uses which password encoder. Then you specify a default one to be used when no prefix is available.

src/main/java/com/faforever/api/data/validation/VotingSubjectRevealWinnerValidator.java
```diff
@@ -0,0 +1,20 @@
...
+
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+import java.time.OffsetDateTime;
+
+public class VotingSubjectRevealWinnerValidator implements ConstraintValidator<VotingSubjectRevealWinnerCheck, VotingSubject> {
+
+  @Override
+  public void initialize(VotingSubjectRevealWinnerCheck constraintAnnotation) {
+    //Comment for codacy to show that this constructor is intentional empty
```
micheljung: Space after `//` 😁

src/main/java/com/faforever/api/config/security/WebSecurityConfig.java
Brutus5000: Yes, I approve the idea in general. But it's not feasible with the current python server, so I would like to postpone this suggestion and keep it simple for now.

PR #256, opened 2018-09-07
Fixes #252

### LINKED ISSUE #252: Improve steam link error messages

1-alex98:


Brutus5000 (changes requested):
Stop removing the throws Exception from the unit tests. This is very annoying to review and it also doesn't make life easier. You add this to all unit tests because you don't care about exceptions (-> the test would fail anyway) but some test operations have checked exceptions like asserts regarding Jackson.

Brutus5000 (dismissed):
The current unit test error indicates we are returning the error json as url parameter. I can't find whether it was introduced by your changes but it can't stay that way.

axel1200:
It was introduced by me I will fix it.
It is my IDE that removes those throws Exceptions. And I don't see a problem there be can add it at those test that need it and leave it out for the others.

Brutus5000 (approved):

PR #261, opened 2018-09-12
Send out additional emails

### LINKED ISSUE #253: Email banned users
Very often users do not know they are banned and what for to make this more transparent we should email banned users and tell them the reason and duration for the ban.

1-alex98:
Fixes #253


Brutus5000:
Before we do this we need to solve #322

PR #268, opened 2018-09-23
Fix travis to push develop as docker image

Brutus5000:
I noticed that develop is not pushed to docker (anymore?).

The `VERSION` variable was derived incorrect on branch pushes. However I don't understand why it worked with tags.
The docker push code was redundant.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ before_install:
   - curl -L https://github.com/docker/compose/releases/download/1.16.1/docker-compose-`uname -s`-`uname -m` > /tmp/docker-compose
   - chmod +x /tmp/docker-compose
   - sudo mv /tmp/docker-compose /usr/local/bin
-  - VERSION=$([ -z "${TRAVIS_TAG}" ] && echo "${TRAVIS_TAG#*v}" || echo "${TRAVIS_BRANCH##*/}")
+  - VERSION=$([ -n "${TRAVIS_TAG}" ] && echo "${TRAVIS_TAG#*v}" || echo "${TRAVIS_BRANCH##*/}")
 
 install:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -30,20 +30,14 @@ after_success:
...
     docker push ${REPO};
     fi
 # TODO Codacy coverage reporter does not yet support Java 9. See https://github.com/codacy/codacy-coverage-reporter/issues/76
 #      ./gradlew sendCoverageToCodacy;
-  - if [ "${TRAVIS_BRANCH}" == "develop" ]; then
-    docker build -t ${IMAGE_TAG} . &&
-    echo "${DOCKER_PASSWORD}" | docker login -u "${DOCKER_USERNAME}" --password-stdin &&
-    docker tag ${IMAGE_TAG} ${REPO}:${VERSION} &&
-    docker push ${REPO};
-    fi
```
micheljung, line 46: I know it's not obvious but the reason why the block is duplicated is because for `develop`, `./gradlew sendCoverageToCodacy;` should be executed (which is currently commented out), but not for tags.
Since codacy/codacy-coverage-reporter#76 is solved, it can be put back into the `if`. If you see a good solution how to remove the redundancy, please feel free to do so :)

Brutus5000: What is the advantage of not sending to coverage on tags? Yes it's redundant, but it's not causing any troubles.

micheljung: Hmm I wouldn't know... maybe I thought it's only relevant (or even working) on `develop` but coveralls supports multibranch (other than SonarQube Community Edition) so we're good 👍

micheljung (approved):

PR #274, opened 2018-10-18
Closes #273 Make maps hidden by default when uploaded

### LINKED ISSUE #273: Make maps hidden by default when uploaded
> MisterK [2:40 PM]
I idea is to create a step for the map maker, so they don't fill the vault will test maps or broken maps or just plain copied maps
I have the team to check every day for maps and test them

bukajsytlos:
Closes #273 Make maps hidden by default when uploaded

### DIFF
--- a/src/main/java/com/faforever/api/map/MapService.java
+++ b/src/main/java/com/faforever/api/map/MapService.java
@@ -268,7 +268,7 @@ public class MapService {
       .setDescription(scenarioInfo.get(ScenarioMapInfo.DESCRIPTION).tojstring().replaceAll("<LOC .*?>", ""))
       .setWidth(size.get(1).toint())
       .setHeight(size.get(2).toint())
-      .setHidden(false)
+      .setHidden(fafApiProperties.getMap().isHiddenByDefault())
       .setRanked(progressData.isRanked())
       .setMaxPlayers(scenarioInfo.get(ScenarioMapInfo.CONFIGURATIONS).get(ScenarioMapInfo.CONFIGURATION_STANDARD).get(ScenarioMapInfo.CONFIGURATION_STANDARD_TEAMS).get(1)
         .get(ScenarioMapInfo.CONFIGURATION_STANDARD_TEAMS_ARMIES).length())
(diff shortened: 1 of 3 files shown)


src/main/java/com/faforever/api/map/MapService.java
```diff
@@ -268,7 +268,7 @@ private void updateMapEntities(MapUploadData progressData) {
       .setDescription(scenarioInfo.get(ScenarioMapInfo.DESCRIPTION).tojstring().replaceAll("<LOC .*?>", ""))
       .setWidth(size.get(1).toint())
       .setHeight(size.get(2).toint())
-      .setHidden(false)
+      .setHidden(true)
```
Brutus5000: Waaait a second... this was hardcoded all the time!? That would explain some reported weird behavior.
Wouldn't any update on this mapVersion make it hidden again? Like unhide it would make it hidden again? So we are in an endless loop?

We need to distinguish between initial creation and update, don't we?

Brutus5000 (changes requested):

1-alex98:
Do they really want to have that feature they are gonna be flooded by maps they need to approve I think that is a mistake doing this. And as permission system is not done the only ppl to be able to approve maps are moderators that is gonna give problems dont you think? I mean it is the m&m teams that says we can approve them but as of now only moderators are able to do that.

I would think of other ways to solve that problem, if you ask me this solution is just the easiest one but not the best one(as I said earlier I would give out permission to upload maps to users that promise they read the rules about map upload/making also you have a rating systems in the vault in downlords client send ppl out to use it from the m&m team and it will get easier to find good maps).

1-alex98:
Also I would advise to change the upload UI of the client at the same time otherwise uploaders will spam #technical-help or forum, not being able to find their maps. Also how would you get back to uploaders to tell them the reason the map was rejected. I think this is really user unfriendly the way it would be after this PR.

src/main/java/com/faforever/api/map/MapService.java
bukajsytlos: do you want to rename that method?

Brutus5000: I checked the code now and not just the changed snippet. It looks fine. Just needs the context.

Brutus5000 (dismissed):

micheljung (changes requested):
Very debatable, and opinions will probably change more than once in future.
Therefore, please make this configurable via `FafApiProperties`.

Brutus5000:
As far as I know this is no longer a requirement.

1-alex98:
If we ever do this. We need a proper way of testing hidden maps and proper way to give out permissions to unhide them, which we should at least have by now...

PR #281, opened 2018-11-01
Ban Revoke structure

### LINKED ISSUE #280: Ban revoke structure change
https://github.com/FAForever/db/pull/176

1-alex98:
Fixes #280


1-alex98:
The correct faf-stack version does not exist yet

src/main/java/com/faforever/api/data/domain/BanInfo.java
```diff
@@ -55,8 +56,8 @@ public Player getPlayer() {
   }
 
   @ManyToOne
+  @UpdatePermission(expression = "Prefab.Role.None")
   @JoinColumn(name = "author_id")
-  @NotNull
```
Brutus5000: Why was @NotNull removed here?

Brutus5000 (changes requested):
Please add integration tests to BanInfoTest.java
They were missing before, because it did not work with Elide. Now as it should be working, we should have tests.

src/main/java/com/faforever/api/data/domain/BanInfo.java
1-alex98: It was removed because it is set now anyway so no @NotNull nedded

Brutus5000: No. Implementations may change, but there is nothing more explicit than validation right on the entity. No reason to remove it. It gives additional information and enforces a validation.

Brutus5000:
When I run this locally on Windows the following tests fail:
com.faforever.api.data.BanTest#canCreateBanInfoBasedOnModerationReportAsModerator
com.faforever.api.data.BanTest#canCreateBanInfoAsModerator
com.faforever.api.data.BanTest#canReadBanInfoAsModerator

Can you check?

Brutus5000:
Looks like it was a local gradle issue with not updating java-commons

Brutus5000 (approved):

PR #283, opened 2018-11-04
Closes #282 Allow login via username or email

### LINKED ISSUE #282: Allow login via username or email
Users have the ability change their usernames frequently. Some of these forget their username after a longer absence and can't login anymore.
It would be more user friendly if users could login with username or email (like in password reset).

bukajsytlos:
Closes #282 Allow login via username or email

### DIFF
--- a/src/main/java/com/faforever/api/security/FafUserDetailsService.java
+++ b/src/main/java/com/faforever/api/security/FafUserDetailsService.java
@@ -26,9 +26,9 @@ public class FafUserDetailsService implements UserDetailsService {
   }
 
   @Override
-  public UserDetails loadUserByUsername(String username) throws UsernameNotFoundException {
-    User user = userRepository.findOneByLoginIgnoreCase(username)
-      .orElseThrow(() -> new UsernameNotFoundException("User could not be found: " + username));
+  public UserDetails loadUserByUsername(String usernameOrEmail) throws UsernameNotFoundException {
+    User user = userRepository.findOneByLoginIgnoreCaseOrEmailIgnoreCase(usernameOrEmail, usernameOrEmail)
+      .orElseThrow(() -> new UsernameNotFoundException("User could not be found: " + usernameOrEmail));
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/faforever/api/security/FafUserDetailsService.java
```diff
@@ -26,9 +26,9 @@ public FafUserDetailsService(UserRepository userRepository) {
   }
 
   @Override
-  public UserDetails loadUserByUsername(String username) throws UsernameNotFoundException {
-    User user = userRepository.findOneByLoginIgnoreCase(username)
-      .orElseThrow(() -> new UsernameNotFoundException("User could not be found: " + username));
+  public UserDetails loadUserByUsername(String usernameOrEmail) throws UsernameNotFoundException {
```
1-alex98, line 29: Dont we still have accounts that have same email and username if compared to lowwer case how would that affect this code

1-alex98 (changes requested):

src/main/java/com/faforever/api/security/FafUserDetailsService.java
bukajsytlos: well good question, but I would argue that emails should be CI and username was searched CI even before. So if there are two accounts with same email, they will fail to log. Let's hope they will contact mod :)

1-alex98 (approved):

src/main/java/com/faforever/api/security/FafUserDetailsService.java
Brutus5000: Ha interesting point: this is also an issue on password reset.
But I have checked, hibernate does not throw an exception.
Two email accounts per user are not wanted, so we will ban of the two users and then need to scrap their email address I guess.

Otherwise: if you have this issue, then just use your login name. All other cases are the user's own fault.

Brutus5000 (approved):

PR #301, opened 2019-01-06
Closes #291 EventService - Updating events fails

### LINKED ISSUE #291: EventService - Updating events fails
When sending a valid request to the `/events/update` route for an existing event uuid an exception is thrown.
**Sample payload:**
```
[{
        'count': 1,
        'playerId': 1,
        'eventId': 'ad193982-e7ca-465c-80b0-5493f9739559'
}]
```

**Exception:** `java.sql.SQLException: Field 'event_id' doesn't have a default value `

**Impact:** Achievements can not be updated from game results.

**First analysis:**
In the EventsService a new `PlayerEvent` is created if not already exists. However the `getEvent` in the `PlayerEvent` is annotated as `insertable = false`. I assume this tells hibernate not to insert it, thus it tries to resolve the default value and fails.
Maybe removing the `insertable = false` already solves the problem

**Acceptance criteria for solution:**
- [ ] Fix the root cause (probably the insertable-flag).
- [ ] Replace `eventRepository.getOne` in the EventsService with `findById` and throw a proper error message if optional is empty
- [ ] Write unit tests for the increment method. Mark the method as `@VisibleForTesting`
- [ ] Write integration tests to actually check MySql behavior (add some achievement test-data into sql fails and preload them - look into other integration tests to find out how to do it)

bukajsytlos:
Closes #291 EventService - Updating events fails

### DIFF
--- a/src/main/java/com/faforever/api/event/EventsService.java
+++ b/src/main/java/com/faforever/api/event/EventsService.java
@@ -2,6 +2,10 @@ package com.faforever.api.event;
 
 import com.faforever.api.data.domain.Event;
 import com.faforever.api.data.domain.PlayerEvent;
+import com.faforever.api.error.ApiException;
+import com.faforever.api.error.Error;
+import com.faforever.api.error.ErrorCode;
+import com.faforever.api.player.PlayerService;
 import com.google.common.base.MoreObjects;
 import org.springframework.stereotype.Service;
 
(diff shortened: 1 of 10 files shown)


louvenarde:
Awesome! <3

src/main/java/com/faforever/api/event/EventsService.java
```diff
@@ -12,17 +16,22 @@
 public class EventsService {
 
   private final EventRepository eventRepository;
+  private final PlayerRepository playerRepository;
```
micheljung: Services should only use repositories of entities for which they are directly responsible for. Therefore, EventsService should use PlayerService here

micheljung (changes requested):
One small change, otherwise it looks rock solid! Many thanks

src/main/java/com/faforever/api/event/EventsService.java
bukajsytlos: done

Brutus5000 (approved):

PR #304, opened 2019-02-03
allow moderators to update steamids

Geosearchef:


1-alex98 (approved):

1-alex98:
U can now merge it

1-alex98:
Oh btw do u want them to allow to change it or only to set it to null? If only to Null look at what I used with the hidden attribute on maps. I made the author only able to set it to true but not to false in a similar way you can only allow to set it to null

Brutus5000 (changes requested):
Audit log annotations are missing

Brutus5000:
Does no longer fit into the permission system.

PR #312, opened 2019-05-05
 Add request id context to logging and error responses

Brutus5000:
Each http request creates a request id (either from nginx or random uuid).
This request id is then applied to every log message during processing until the http request is finished, then it is cleaned up.
Furthermore the request id is added to each error message the API sends back to the client. So the user has an id that he can give on error reports to check the related logs.

### DIFF
--- a/src/main/java/com/faforever/api/error/ErrorResponse.java
+++ b/src/main/java/com/faforever/api/error/ErrorResponse.java
@@ -1,12 +1,20 @@
 package com.faforever.api.error;
 
+import com.faforever.api.logging.RequestIdFilter;
 import lombok.Data;
+import org.slf4j.MDC;
 
 import java.util.ArrayList;
 
 @Data
 public class ErrorResponse {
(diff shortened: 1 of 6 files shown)


src/main/java/com/faforever/api/error/ErrorResponse.java
```diff
@@ -1,13 +1,21 @@
...
 
+import com.faforever.api.logging.RequestIdFilter;
 import lombok.Data;
+import org.slf4j.MDC;
 
 import java.util.ArrayList;
 
 @Data
 public class ErrorResponse {
+  private final String requestId;
```
bukajsytlos, line 11: why dont we leave it as response header?

Brutus5000: That's probably my personal preference. As part of the payload I have it immediately available e.g. when deserializing the message with a Spring RestTemplate

PR #323, opened 2019-07-16
Replace the MacSigner with RsaSigner/RsaVerifier. (fixes #318)

### LINKED ISSUE #318: Use asymmetric JWS in auth tokens
We wish to support authentication via API tokens in the server, so that passwords are handled only by the API. To do this we would like for the server to be able to verify the signature of an API token which currently would require the server to know the API secret used to sign tokens (meaning the server could sign it's own API tokens). However, if API token signatures were to use an asymmetric key variant of JWS (such as RS256), then the server could verify that a token was signed by the API with only the API's ***public*** key, avoiding the problem of the server being able to sign it's own API tokens.

See the server discussion here:
https://github.com/FAForever/server/issues/440

Brutus5000:
This allows us to hand out the public key to other services. Especially
the JWT needs to be verified by the lobby server to get rid of multiple
services accessing the user passwords.


1-alex98 (dismissed):

Brutus5000:
Need to figure out, why the integration tests fail :(

1-alex98 (approved):

PR #324, opened 2019-07-25
Upgraded gradle version to 5.5.1

breucode:


gradle.properties
```diff
@@ -31,3 +31,4 @@ jsonapiConverterVersion=0.9
 codacyCoverageReporterVersion=4.0.0
 jsonVersion=20180813
 javaxInterceptorApiVersion=1.2
+lombokVersion:1.18.4
```
Brutus5000: I guess you need an equal sign here?

src/lombok.config
```diff
@@ -1,3 +1,4 @@
 lombok.anyConstructor.addConstructorProperties=true
 lombok.accessors.chain=true
 config.stopBubbling=true
+lombok.addLombokGeneratedAnnotation = true
```
Brutus5000: Remove whitespaces

Brutus5000 (changes requested):
Cool, thanks a lot. Some minor stuff though.

PR #342, opened 2019-09-23
Implement user permission system

Brutus5000:


.travis.yml
```diff
@@ -23,7 +23,7 @@ before_install:
 install:
   - git clone https://github.com/FAForever/faf-stack.git faf-stack
     && pushd faf-stack
-    && git checkout e33243c9
+    && git checkout 79c5d9d9
```
norraxx: Maybe let's start using branch names? At least it will be visible what commits do we need.
It's propositon to make things better  :-)

Brutus5000: Branch names aren't fixed. You could not rerun and older pipeline if the branch (e.g. develop) has changed over time.

louvenarde: (Maybe with tags then?)

Brutus5000: Well you can use tags already, but the problem is the development phase when there is no tag yet.

PR #328, opened 2019-08-05
#327 Exe uploads

norraxx:
Don't merge, just for review purposes

PR #345, opened 2019-09-24
Frontend for exe uploads

norraxx:
Frontend for api: https://github.com/FAForever/faf-java-api/pull/328


src/main/java/com/faforever/api/deployment/ExeUploaderController.java
```diff
@@ -32,6 +36,11 @@ public ExeUploaderController(
     this.exeUploaderService = exeUploaderService;
   }
 
+  @RequestMapping(path = "/upload_form", method = RequestMethod.GET, produces = TEXT_HTML_VALUE)
```
Brutus5000: I would rename this to just /upload . GET gives the form, POST does the action.

Brutus5000 (changes requested):
Solid PR, just one minor change request 😄

src/main/java/com/faforever/api/deployment/ExeUploaderController.java
norraxx: Np

Issue #348, opened 2019-09-29
Code cleanup of Spring components


Brutus5000:
- [ ] Remove all `@Inject` and `@Autowired` annotations from the main modules (not the tests!). They are obsolete and no longer required.
- [ ] Replace hand-written constructors for these components with Lombok's `@AllArgsConstructor` or `@RequiredArgsConstructor` (where applicable - does not work in rare cases where the constructor is doing more than just assigning `private final` variables)
- [ ] Replace `@PostConstruct` and `@PreDestroy` methods -> add the interface implementations for `InitializingBean` and `DisposableBean` and move the code to the interface method.

vaibhavthapliyal:
I would like to work on this issue.

PR #350, opened 2019-10-02
#348 cleanup if Spring components

marcoebbinghaus:
I did everything that was mentioned for that issue. The classes that still have an @Inject annotation can't be replaced trivially.

### DIFF
--- a/src/main/java/com/faforever/api/achievements/AchievementService.java
+++ b/src/main/java/com/faforever/api/achievements/AchievementService.java
@@ -7,6 +7,7 @@ import com.faforever.api.data.domain.PlayerAchievement;
 import com.faforever.api.error.ApiException;
 import com.faforever.api.error.Error;
 import com.google.common.base.MoreObjects;
+import lombok.RequiredArgsConstructor;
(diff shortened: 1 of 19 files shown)


Brutus5000 (commented):
Can you explain why / where you think replacing the constructor with `@RequiredArgsConstructor` is not possible?
As far as I can see the classes
- ClanService
- AchievementService
- ClansController
- FeaturedModEnricher
do not do anything special in the constructor, so it should work here.

marcoebbinghaus:
AchievementService: I think I just missed to replace the constructor here, sorry for that.
ClanService: The ClanServiceTest made usage of the constructor. But I removed the constructor and aligned the test now so it works without the constructor.
ClansController: I did replace the constructor here, what exactly do you miss?
FeaturedModEnricher: FeaturedModEnricher does not have a constructor and it only has a static field. I think Lombok annotations do not work with static fields ("All static fields will be skipped entirely (not considered for any of the generated methods, and no setter/getter will be made for them)." from https://projectlombok.org/features/Data).

I pushed the according changes.

Brutus5000 (approved):

PR #351, opened 2019-10-05
WIP: Migrate all @Entity annotations from property to field level

### LINKED ISSUE #346: Migrate all @Entity annotations (columns and relations) from method level to field level
The Elide library should support field-level hibernate annotations for quite a while now.

Since we use Lombok, creating the Getter-methods manually is unnecessary code and I would like to move the annotation to the field level variables.

In case new problems occur, we'd need to create a bug ticket at elide bug tracker (https://github.com/yahoo/elide/issues).

vanogrid:
Closes #346 

Migrate all entity annotations from property to field level and use Lombok for the Getters.

WIP: After moving the annotations there are 8 failing integration tests. The issue seems to be related to the lazy-fetched collections (e.g. UserNote.java). The tests passed when I replaced `repository.getOne(id)` with `repository.findById(id)` inside the tests and changed the fetch strategy of some collections from Lazy to Eager. I think Elide has a problem using Hibernate Proxy objects when field access type is used, but I was able to set up a new Spring Boot and Elide project (without Spring Security) and I couldn't reproduce the problem, so I can't confirm yet.

This [issue](https://github.com/yahoo/elide/issues/844) might be related, but I can't confirm for sure as the id returned in our case is `0`, not `null`, and we're having problems mostly with the ManyToOne relationships.

### DIFF
--- a/src/main/java/com/faforever/api/client/OAuthClient.java
+++ b/src/main/java/com/faforever/api/client/OAuthClient.java
@@ -1,5 +1,7 @@
 package com.faforever.api.client;
 
+import lombok.AccessLevel;
+import lombok.Getter;
 import lombok.Setter;
 import org.jetbrains.annotations.Nullable;
 import org.springframework.util.Assert;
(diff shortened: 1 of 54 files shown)


Brutus5000:
I have encountered this error a few weeks ago. Unfortunately I do not know what causes it. Could be anything from Hibernate caching to transaction issues etc.
My approach was not to use the JpaRepositories in the tests, but to query for the object using mockMvc and asserting stuff with jsonPath-evaluation.

If you find a solution to the Hibernate problem it would be awesome, but I would also be happy with evading the problem using mockMvc. If you need help with that feel free to ping me.

vanogrid:
Thanks! I will see what I can do :muscle:

Brutus5000:
@vanogrid Any news? Or should I take over here?

vanogrid:
@Brutus5000 Sorry, I could not fix the issue. You can take over. I still believe it is related to Elide and the way it maps the entity when using field access, but I wasn't able to confirm. 

When the object is not a proxy, it works fine, but if it is, the field is null. I tried to make Elide map the field to the Getter method, so the proxy can be initialized when called, but I couldn't make it work. I guess there should be some logic in Elide to initialize the proxy beforehand. Then the mapping might work correctly. Using only mockMvc is also not an option, because there are some tests which use mockMvc already, but they test relationships and the related objects are null, so the tests keep failing.

I will close the PR, as it is outdated and causes a lot of merge conflicts when rebased.

PR #352, opened 2019-10-21
Migrate tests to junit 5

icougil:


Brutus5000:
The broken pipeline seems to be caused by Travis CI changes. For some reason the JDK11 compiler is used while it was JDK8 a few days ago 😦

icougil:
About the build, it seems [Oracle JDK 9 is discontinued in Travis CI](https://docs.travis-ci.com/user/languages/java/#using-java-10-and-later). 
Probably you should define the JDK version required to compile the artifacts and see it the build works (+ also think about not using the proprietary Oracle JDK version and try to move to OpenJDK 😉  )

Brutus5000:
If you add `dist: trusty` as a top line to the .travis.yml we should be back to a working state. I know it's a dirty hack, but right now I have no time to migrate to OpenJDK11 (which would be my preferred step for now).
If you want to give me a hand, that would be highly appreciated 😄

Brutus5000 (approved):

PR #366, opened 2019-11-22
Add support for user data updates forwarding to NodeBB

Brutus5000:


src/main/java/com/faforever/api/forum/NodebbService.java
```diff
@@ -0,0 +1,102 @@
...
+  private final RestTemplate restTemplate;
+  private final FafApiProperties properties;
+
+  @Override
+  public void userDataChanged(UserUpdatedEvent event) {
+    try {
+      getNodebbUserId(event.getId())
+        .ifPresentOrElse(
+          userId -> updateUserData(userId, event),
+          () -> log.info("User data not updated in NodeBB (User not found): {}", event)
```
1-alex98, line 44: warn? Becuase shold not happen

1-alex98 (dismissed):

src/main/java/com/faforever/api/forum/NodebbService.java
Brutus5000: You only have a user account in NobdeBB after you first logged in via SSO. So 99% of our userbase will not have a user.

src/test/java/com/faforever/api/user/UserServiceTest.java
```diff
@@ -95,6 +96,8 @@
   private GlobalRatingRepository globalRatingRepository;
   @Mock
   private Ladder1v1RatingRepository ladder1v1RatingRepository;
+  @Mock
+  private ApplicationEventPublisher eventPublisher;
```
bukajsytlos, line 97: no assert on event publishing and its content?

Brutus5000: Fixed now :)

src/main/java/com/faforever/api/mautic/MauticService.java
```diff
@@ -56,22 +62,27 @@ public MauticService(MappingJackson2HttpMessageConverter mauticApiMessageConvert
...
+      "faf_user_id", event.getId(),
+      "faf_username", event.getUsername()
+    );
 
-    return CompletableFuture.completedFuture(restOperations.postForObject("/contacts/new", body, Object.class));
+    try {
+      restTemplate.postForObject("/contacts/new", body, Object.class);
+      log.debug("Updated contact in Mautic from event: {}", event);
+    } catch (Exception e) {
+      log.error("Could not update contact in Mautic for user id: {}", event.getId(), e);
```
1-alex98, line 85: error is supposed to be logged if application gets unfunctional by error, so maybe this would might rather be warn(maybe mautic is down)...

src/main/java/com/faforever/api/forum/NodebbService.java
```diff
@@ -0,0 +1,108 @@
...
+import org.springframework.util.LinkedMultiValueMap;
+import org.springframework.web.client.HttpClientErrorException;
+import org.springframework.web.client.RestTemplate;
+import org.springframework.web.util.UriComponentsBuilder;
+
+import java.net.URI;
+import java.util.Optional;
+
+@Service
+@ConditionalOnProperty(value = "faf-api.nodebb.master-token")
```
1-alex98, line 26: So does Spring not create this bean if that  value is set to "false"

1-alex98 (dismissed):
Looks solid to me

src/main/java/com/faforever/api/mautic/MauticService.java
Brutus5000: This is exactly the point. If Mautic is down, faf and mautic go out of sync , which is an error and should be fixed (there is no way to do it right now).

src/main/java/com/faforever/api/forum/NodebbService.java
Brutus5000: Yes.

PR #388, opened 2020-03-29
Closes #386 Fix filtering based on enums

### LINKED ISSUE #386: Fix filtering based on enums
https://github.com/FAForever/faf-java-api/pull/384

bukajsytlos:
Looks like the only broken enum left was faction

### DIFF
--- a/src/main/java/com/faforever/api/data/domain/Faction.java
+++ b/src/main/java/com/faforever/api/data/domain/Faction.java
@@ -1,8 +1,5 @@
 package com.faforever.api.data.domain;
 
-import com.fasterxml.jackson.annotation.JsonCreator;
-import com.fasterxml.jackson.annotation.JsonValue;
-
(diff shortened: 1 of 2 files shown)


Brutus5000 (dismissed):

src/main/java/com/faforever/api/data/domain/Faction.java
```diff
@@ -26,7 +23,6 @@
     this.string = string;
   }
 
-  @JsonCreator
```
Brutus5000, line 29: Why removed?

bukajsytlos: well we said all string, so we want to send it as string and receive it as string

Brutus5000: Ok. Since we don't parse any replays it should be fine.

PR #392, opened 2020-04-16
Add checks for vault bans on uploading maps and mods (fixes #380)

### LINKED ISSUE #380: Give M&M and Vault Maintenance Members Ability to Ban Vault Usage
After some discussion with the Moderation councilor, we believe that a client ban on users for abusing the vaults and violating the rules is too harsh. A user that causes issue with the client vault should not be prohibited from playing the game, unless moderation believes they are directly affecting rulings on general FAF usage.

If possible, we would like for the ability to be given so that M&M members with access to mordor can prohibit users from uploading to vaults after repeat, or extreme, negligence to the rules.

Brutus5000:
Error response looks like this:

```
{
    "requestId": "6887b2c7-130d-4267-a064-be6dff37cb92",
    "errors": [
        {
            "status": "403",
            "title": "You are not allowed to access this resource.",
            "detail": "You are permanently banned from uploading mods to the vault."
        }
    ]
}
```

### DIFF
--- a/src/main/java/com/faforever/api/map/MapService.java
+++ b/src/main/java/com/faforever/api/map/MapService.java
@@ -202,9 +202,7 @@ public class MapService {
   }
 
   private void checkAuthorVaultBan(Player author) {
-    author.getActiveBans().stream()
-      .filter(ban -> ban.getLevel() == BanLevel.VAULT)
-      .findFirst()
+    author.getActiveBanOf(BanLevel.VAULT)
       .ifPresent((banInfo) -> {
         String message = banInfo.getDuration() == BanDurationType.PERMANENT ?
           "You are permanently banned from uploading maps to the vault." :
--- a/src/main/java/com/faforever/api/map/MapService.java
+++ b/src/main/java/com/faforever/api/map/MapService.java
@@ -2,6 +2,8 @@ package com.faforever.api.map;
 
 import com.faforever.api.config.FafApiProperties;
 import com.faforever.api.content.ContentService;
+import com.faforever.api.data.domain.BanDurationType;
+import com.faforever.api.data.domain.BanLevel;
 import com.faforever.api.data.domain.Map;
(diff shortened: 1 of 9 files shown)


src/main/java/com/faforever/api/map/MapService.java
```diff
@@ -194,6 +201,19 @@ public void uploadMap(InputStream mapDataInputStream, String mapFilename, Player
     }
   }
 
+  private void checkAuthorVaultBan(Player author) {
+    author.getActiveBans().stream()
+      .filter(ban -> ban.getLevel() == BanLevel.VAULT)
+      .findFirst()
```
bukajsytlos: I would put this into player class

src/test/java/com/faforever/api/mod/ModServiceTest.java
```diff
@@ -110,6 +115,21 @@ public void testExistingUid() throws Exception {
...
+  public void testUploaderVaultBanned() throws Exception {
+    Path uploadedFile = prepareMod(TEST_MOD);
+
+    Player uploader = mock(Player.class);
+    when(uploader.getActiveBans()).thenReturn(Set.of(
+      new BanInfo()
+        .setLevel(BanLevel.VAULT)
+    ));
+
+    expectedException.expect(Forbidden.class);
```
bukajsytlos, line 127: why different approach here?

Brutus5000: Not sure, which approach you are referring to but:

a) We can't use the matcher as it is no ApiException.
b) I don't want an ApiException, as this would not cause an 403 status code.
c) The test is still JUnit 4 codebase, so I did not bother to migrate it to JUnit5 assertThrows.

src/main/java/com/faforever/api/map/MapService.java
Brutus5000: Done

PR #401, opened 2020-06-18
Add Filtering to Leaderboard

### LINKED ISSUE #400: Add a way to filter in the leaderboard controller
Provide regex via query param to filter list and add meta data about total number of pages

1-alex98:
Fixes #400

### DIFF
--- a/src/inttest/java/com/faforever/api/leaderboard/LeaderboardControllerTest.java
+++ b/src/inttest/java/com/faforever/api/leaderboard/LeaderboardControllerTest.java
@@ -23,7 +23,7 @@ public class LeaderboardControllerTest extends AbstractIntegrationTest {
 
     mockMvc.perform(asyncDispatch(mvcResult))
       .andExpect(status().isOk())
-      .andExpect(content().string("{\"data\":[{\"type\":\"ladder1v1LeaderboardEntry\",\"id\":\"1\",\"attributes\":{\"name\":\"USER\",\"mean\":1500.0,\"deviation\":120.0,\"numGames\":5,\"wonGames\":1,\"rank\":1,\"rating\":1140}},{\"type\":\"ladder1v1LeaderboardEntry\",\"id\":\"2\",\"attributes\":{\"name\":\"MODERATOR\",\"mean\":1200.0,\"deviation\":90.0,\"numGames\":5,\"wonGames\":2,\"rank\":2,\"rating\":930}},{\"type\":\"ladder1v1LeaderboardEntry\",\"id\":\"3\",\"attributes\":{\"name\":\"ADMIN\",\"mean\":1000.0,\"deviation\":100.0,\"numGames\":5,\"wonGames\":3,\"rank\":3,\"rating\":700}}]}"));
+      .andExpect(content().string("{\"data\":[{\"type\":\"ladder1v1LeaderboardEntry\",\"id\":\"1\",\"attributes\":{\"name\":\"USER\",\"mean\":1500.0,\"deviation\":120.0,\"numGames\":5,\"wonGames\":1,\"rank\":1,\"rating\":1140}},{\"type\":\"ladder1v1LeaderboardEntry\",\"id\":\"2\",\"attributes\":{\"name\":\"MODERATOR\",\"mean\":1200.0,\"deviation\":90.0,\"numGames\":5,\"wonGames\":2,\"rank\":2,\"rating\":930}},{\"type\":\"ladder1v1LeaderboardEntry\",\"id\":\"3\",\"attributes\":{\"name\":\"ADMIN\",\"mean\":1000.0,\"deviation\":100.0,\"numGames\":5,\"wonGames\":3,\"rank\":3,\"rating\":700}}],\"meta\":{\"totalPages\":1,\"totalRecords\":3}}"));
   }
 
   @Test
@@ -33,7 +33,7 @@ public class LeaderboardControllerTest extends AbstractIntegrationTest {
 
     mockMvc.perform(asyncDispatch(mvcResult))
       .andExpect(status().isOk())
-      .andExpect(content().string("{\"data\":[{\"type\":\"ladder1v1LeaderboardEntry\",\"id\":\"1\",\"attributes\":{\"name\":\"USER\",\"mean\":1500.0,\"deviation\":120.0,\"numGames\":5,\"wonGames\":1,\"rank\":1,\"rating\":1140}}]}"));
+      .andExpect(content().string("{\"data\":[{\"type\":\"ladder1v1LeaderboardEntry\",\"id\":\"1\",\"attributes\":{\"name\":\"USER\",\"mean\":1500.0,\"deviation\":120.0,\"numGames\":5,\"wonGames\":1,\"rank\":1,\"rating\":1140}}],\"meta\":{\"totalPages\":3,\"totalRecords\":3}}"));
   }
 
   @Test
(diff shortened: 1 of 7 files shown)


src/main/java/com/faforever/api/leaderboard/LeaderboardController.java
```diff
@@ -32,12 +35,22 @@ public LeaderboardController(LeaderboardService leaderboardService) {
...
   @Async
   @RequestMapping(path = "/ladder1v1", method = RequestMethod.GET)
   @ApiOperation("Lists the ladder1v1 leaderboard")
-  public CompletableFuture<JsonApiDocument> getLadder1v1(@RequestParam(value = "page[number]", required = false) Integer page,
-                                                         @RequestParam(value = "page[size]", required = false) Integer pageSize) {
-    List<Resource> values = StreamSupport.stream(leaderboardService.getLadder1v1Leaderboard(page, pageSize).spliterator(), false)
+  public CompletableFuture<JsonApiDocument> getLadder(@RequestParam(value = "page[number]", required = false) Integer page,
+                                                      @RequestParam(value = "page[size]", required = false) Integer pageSize,
+                                                      @ApiParam("Only players that have the following pattern in their name.")
+                                                      @RequestParam(value = "playerNameMatchesRegex", required = false) String playerNameMatchesRegex) {
```
bukajsytlos, line 45: playerNameMatchesRegex is very confusing/wrong name, it's not regex and Matches suggests its boolean.
I would leave just playerName or playerNamePattern

src/main/java/com/faforever/api/leaderboard/LeaderboardController.java
```diff
@@ -32,12 +35,22 @@ public LeaderboardController(LeaderboardService leaderboardService) {
...
+    Page<Ladder1v1LeaderboardEntry> ladder1v1LeaderboardEntries;
+    if(playerNameMatchesRegex == null){
+      ladder1v1LeaderboardEntries = leaderboardService.getLadder1v1Leaderboard(page, pageSize);
+    }else {
```
bukajsytlos, line 49: formatting

src/main/java/com/faforever/api/leaderboard/LeaderboardController.java
```diff
@@ -32,12 +35,22 @@ public LeaderboardController(LeaderboardService leaderboardService) {
...
   @ApiOperation("Lists the ladder1v1 leaderboard")
-  public CompletableFuture<JsonApiDocument> getLadder1v1(@RequestParam(value = "page[number]", required = false) Integer page,
-                                                         @RequestParam(value = "page[size]", required = false) Integer pageSize) {
-    List<Resource> values = StreamSupport.stream(leaderboardService.getLadder1v1Leaderboard(page, pageSize).spliterator(), false)
+  public CompletableFuture<JsonApiDocument> getLadder(@RequestParam(value = "page[number]", required = false) Integer page,
+                                                      @RequestParam(value = "page[size]", required = false) Integer pageSize,
+                                                      @ApiParam("Only players that have the following pattern in their name.")
+                                                      @RequestParam(value = "playerNameMatchesRegex", required = false) String playerNameMatchesRegex) {
+    Page<Ladder1v1LeaderboardEntry> ladder1v1LeaderboardEntries;
+    if(playerNameMatchesRegex == null){
```
bukajsytlos, line 47: or isEmpty()

src/main/java/com/faforever/api/leaderboard/GlobalLeaderboardRepository.java
```diff
@@ -23,14 +23,40 @@
...
+    "    global_rating.numGames," +
+    "    @s \\:= @s + 1 rank" +
+    "  FROM global_rating JOIN login on login.id = global_rating.id," +
+    "    (SELECT @s \\:= ?#{#pageable.offset}) AS s" +
+    "  WHERE is_active = 1 AND global_rating.numGames > 0" +
+    "   AND login.id NOT IN (" +
+    "     SELECT player_id FROM ban" +
+    "     WHERE (expires_at is null or expires_at > NOW()) AND revoke_time IS NULL" +
+    "  ) " +
+    "   AND login.login LIKE ?#{#playerNameRegex}"+
```
bukajsytlos, line 48: interesting that this works. but I would prefer `:playerNameRegex`

bukajsytlos (changes requested):

src/main/java/com/faforever/api/leaderboard/LeaderboardService.java
```diff
@@ -30,11 +30,21 @@ public LeaderboardService(GlobalLeaderboardRepository globalLeaderboardRepositor
     return ladder1v1LeaderboardRepository.getLeaderboardByPage(getPageable(page, pageSize));
   }
 
+  @Cacheable(LEADERBOARD_RANKED_1V1_CACHE_NAME)
```
Brutus5000, line 33: I don't think caching makes sense here. Also if it does I don't think you should use the same for the cache key here.

src/main/java/com/faforever/api/leaderboard/LeaderboardService.java
```diff
@@ -30,11 +30,21 @@ public LeaderboardService(GlobalLeaderboardRepository globalLeaderboardRepositor
...
+  public Page<Ladder1v1LeaderboardEntry> getLadder1v1LeaderboardAndFilterPlayerByRegex(@Nullable Integer page, @Nullable Integer pageSize, @NotNull String playNameRgex) {
```
Brutus5000, line 34: typo in Regex

src/main/java/com/faforever/api/leaderboard/LeaderboardService.java
```diff
@@ -30,11 +30,21 @@ public LeaderboardService(GlobalLeaderboardRepository globalLeaderboardRepositor
...
+    return ladder1v1LeaderboardRepository.getLeaderboardByPageAndPlayerNameMatchesRegex(getPageable(page, pageSize), playNameRgex);
+  }
+
   @Cacheable(LEADERBOARD_GLOBAL_CACHE_NAME)
   public Page<GlobalLeaderboardEntry> getGlobalLeaderboard(@Nullable Integer page, @Nullable Integer pageSize) {
     return globalLeaderboardRepository.getLeaderboardByPage(getPageable(page, pageSize));
   }
 
+  @Cacheable(LEADERBOARD_GLOBAL_CACHE_NAME)
```
Brutus5000, line 43: I don't think caching makes sense here. Also if it does I don't think you should use the same for the cache key here.

src/main/java/com/faforever/api/leaderboard/LeaderboardService.java
```diff
@@ -30,11 +30,21 @@ public LeaderboardService(GlobalLeaderboardRepository globalLeaderboardRepositor
...
+  public Page<GlobalLeaderboardEntry> getGlobalLeaderboardAndFilterPlayerByRegex(@Nullable Integer page, @Nullable Integer pageSize, @NotNull String playNameRgex) {
```
Brutus5000, line 44: typo in Regex

src/inttest/java/com/faforever/api/leaderboard/LeaderboardControllerTest.java
```diff
@@ -23,7 +23,7 @@ public void ladder1v1All() throws Exception {
...
+      .andExpect(content().string("{\"data\":[{\"type\":\"ladder1v1LeaderboardEntry\",\"id\":\"1\",\"attributes\":{\"name\":\"USER\",\"mean\":1500.0,\"deviation\":120.0,\"numGames\":5,\"wonGames\":1,\"rank\":1,\"rating\":1140}},{\"type\":\"ladder1v1LeaderboardEntry\",\"id\":\"2\",\"attributes\":{\"name\":\"MODERATOR\",\"mean\":1200.0,\"deviation\":90.0,\"numGames\":5,\"wonGames\":2,\"rank\":2,\"rating\":930}},{\"type\":\"ladder1v1LeaderboardEntry\",\"id\":\"3\",\"attributes\":{\"name\":\"ADMIN\",\"mean\":1000.0,\"deviation\":100.0,\"numGames\":5,\"wonGames\":3,\"rank\":3,\"rating\":700}}],\"meta\":{\"totalPages\":1,\"totalRecords\":3}}"));
```
Brutus5000, line 26: Don't assert json via String. The order should not take any role.
Check MockMvcResultMatchers.jsonPath instead to check each property.
(Applies to all tests)

Brutus5000 (changes requested):

src/main/java/com/faforever/api/leaderboard/GlobalLeaderboardRepository.java
1-alex98: that does not work. U are supposed to use one ways to reference parameters only. I go for Spel type of reference. Your suggestion does not work... I tried

src/main/java/com/faforever/api/leaderboard/LeaderboardController.java
1-alex98: playerNamePattern it is

src/main/java/com/faforever/api/leaderboard/LeaderboardController.java
1-alex98: yeah

src/main/java/com/faforever/api/leaderboard/LeaderboardController.java
1-alex98: yes

src/main/java/com/faforever/api/leaderboard/GlobalLeaderboardRepository.java
bukajsytlos: thanks good to know

bukajsytlos, line 48: now that I think of, is this really what we want to achieve? original query relies on all players to be taken into account to set proper rank. now if you filter by name, you will rank from 1 all matching players, which I suppose is not the goal.
Overall shouldn't we change this to materialized view refreshed every few minutes and use regular elide mapping?
@Brutus5000 @norraxx ?

bukajsytlos (changes requested):

src/main/java/com/faforever/api/leaderboard/GlobalLeaderboardRepository.java
1-alex98: U are right there

1-alex98: Not sure if @Brutus5000 is ok with that

1-alex98: So I would know how to correct it... just make the +1 part in a select

Brutus5000: We have a scheduled job already to purge the leaderboards for inactive players. We could as well make a scheduled job to rebuild ranked leaderboards each time (using MySQL events it is the closest to a materialized view we can get)

If we can do that, we can pull it into Elide.

1-alex98: 🔝 will try one day

Brutus5000:
Closed in favour of #405

PR #405, opened 2020-06-24
Add Filtering to Leaderboard

### LINKED ISSUE #400: Add a way to filter in the leaderboard controller

1-alex98:
Fixes #400


Brutus5000 (dismissed):
I will perform some tests on test server before merging

src/main/java/com/faforever/api/data/domain/GlobalRatingWithRank.java
```diff
@@ -0,0 +1,12 @@
...
+
+import com.yahoo.elide.annotation.Include;
+
+import javax.persistence.Entity;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "global_rating_rank_view")
+@Include(rootLevel = true, type = "globalRatingWithRank")
+public class GlobalRatingWithRank extends RatingWithRank {
```
bukajsytlos: shouldn't we add some read only permission?

bukajsytlos:
also integration tests should be possible right?

Brutus5000:
Actually there is no sense in keeping the old rating and having 2 rating entities

src/main/java/com/faforever/api/data/domain/GlobalRatingWithRank.java
Brutus5000: No by default its read for everybody

Brutus5000:
@bukajsytlos I would add integration tests only if there are permissions or other logic involved.

Brutus5000 (approved):

Issue #363, opened 2019-11-13
Map validation issues


Brutus5000:
The next version also show the latest version. Add `+1`
If you press enter insteaf of clicking a button on the page you get an internal server error.

PR #416, opened 2020-09-08
#363 Increment the next version

ahsanbagwan:
The next version incremented if this is the ask.

### DIFF
--- a/src/main/java/com/faforever/api/map/MapService.java
+++ b/src/main/java/com/faforever/api/map/MapService.java
@@ -95,9 +95,10 @@ public class MapService {
     Integer nextVersion = mapRepository.findOneByDisplayName(mapNameBuilder.getDisplayName())
       .map(map -> map.getVersions().stream()
         .mapToInt(MapVersion::getVersion)
+        .map(i -> i + 1)
         .max()
         .orElse(1))
-      .orElse(2);
+      .orElse(1);
 
     return MapNameValidationResponse.builder()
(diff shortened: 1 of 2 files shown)


Brutus5000:
The endpoint for validation can also be used by other clients directly. I'd rather fix the map service, that provides the non-incremented map version. The html is just one frontend.

ahsanbagwan:
@Brutus5000 Thanks for reviewing this PR.

I'm not sure if this will suffice but at this point adding the number `2` to the outer `orElse` seems right. 

```
Integer nextVersion = mapRepository.findOneByDisplayName(mapNameBuilder.getDisplayName())
      .map(map -> map.getVersions().stream()
        .mapToInt(MapVersion::getVersion)
        .max()
        .orElse(1))
      .orElse(2);
```

src/main/java/com/faforever/api/map/MapService.java
```diff
@@ -97,7 +97,7 @@ public MapNameValidationResponse requestMapNameValidation(String mapName) {
         .mapToInt(MapVersion::getVersion)
         .max()
         .orElse(1))
-      .orElse(1);
+      .orElse(2);
```
Brutus5000: the +1 must be applied on the max value `.map(i -> i+1)`

The .orElse part covers the case, that there is no existing map Version, in that case 1 is the next version

Brutus5000 (approved):

PR #427, opened 2020-11-27
Fixes #426 username reservation check

### LINKED ISSUE #426: Username check for reserved username is broken
As it turns out the check wether a username is reserved after renaming is broken.
https://github.com/FAForever/faf-java-api/blob/5ceaa8ca39d0ac2a7416d28becf6c43ab84931a9/src/main/java/com/faforever/api/user/NameRecordRepository.java#L15

`now()` should be smaller than `change date + x month` if a username is still reserved

Brutus5000:


src/main/java/com/faforever/api/user/NameRecordRepository.java
```diff
@@ -12,6 +12,6 @@
   @Query(value = "SELECT datediff(now(), change_time) FROM name_history WHERE user_id = :userId AND datediff(now(), change_time) <= :maximumDaysAgo ORDER BY change_time DESC LIMIT 1", nativeQuery = true)
   Optional<BigInteger> getDaysSinceLastNewRecord(@Param("userId") Integer userId, @Param("maximumDaysAgo") Integer maximumDaysAgo);
 
-  @Query(value = "SELECT user_id FROM name_history WHERE previous_name = :name AND now() > date_add(change_time, INTERVAL :months MONTH) ORDER BY change_time DESC LIMIT 1", nativeQuery = true)
+  @Query(value = "SELECT user_id FROM name_history WHERE previous_name = :name AND now() < date_add(change_time, INTERVAL :months MONTH) ORDER BY change_time DESC LIMIT 1", nativeQuery = true)
```
bukajsytlos: as I said in the issue I think it would be better to do `now() - x month < change_date` if we use or will be using index on this column

bukajsytlos (changes requested):

src/main/java/com/faforever/api/user/NameRecordRepository.java
Brutus5000: Yes, good suggestion.

PR #435, opened 2021-02-10
Implement support for tokens from FAFs Ory Hydra

### LINKED ISSUE #434: Support OAuth tokens from Ory Hydra along with legacy tokens
The Spring Security OAuth library is deprecated and the replacement does not offer the features of an OAuth Authorisation server.
Therefore reimplemented OAuth 2.0 OpenID Connect login for FAF in the faf-user-service. However since we can't have a big-bang migration we need to have a transition path.

Therefore we need to support the current tokens as well as the new customized Ory Hydra OpenID Connect tokens.

Brutus5000:
Closes #434

### DIFF
--- a/src/main/java/com/faforever/api/security/FafMultiTokenStore.java
+++ b/src/main/java/com/faforever/api/security/FafMultiTokenStore.java
@@ -1,0 +1,112 @@
+package com.faforever.api.security;
+
+import com.faforever.api.config.FafApiProperties;
+import org.springframework.security.oauth2.common.OAuth2AccessToken;
+import org.springframework.security.oauth2.common.OAuth2RefreshToken;
+import org.springframework.security.oauth2.provider.OAuth2Authentication;
+import org.springframework.security.oauth2.provider.token.TokenStore;
+import org.springframework.security.oauth2.provider.token.store.JwtAccessTokenConverter;
+import org.springframework.security.oauth2.provider.token.store.JwtTokenStore;
+import org.springframework.security.oauth2.provider.token.store.jwk.JwkTokenStore;
+import org.springframework.stereotype.Component;
+
+import java.util.Collection;
+import java.util.Collections;
+
+@Component
+public class FafMultiTokenStore implements TokenStore {
+  /**
+   * The token store for token created by this API (using custom OAuth 2.0 without OpenID Connect)
+   */
+  private final JwtTokenStore classicTokenStore;
+
+  /**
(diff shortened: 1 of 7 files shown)


src/main/java/com/faforever/api/security/FafMultiTokenStore.java
```diff
@@ -0,0 +1,112 @@
...
+  @Override
+  public OAuth2Authentication readAuthentication(OAuth2AccessToken token) {
+    return readAuthentication(token.getValue());
+  }
+
+  @Override
+  public OAuth2Authentication readAuthentication(String token) {
+    try {
+      return classicTokenStore.readAuthentication(token);
+    } catch (Exception e) {
```
bukajsytlos: I guess it is too much effort to use more specific exceptions?

src/main/resources/config/application-dev.yml
```diff
@@ -3,6 +3,7 @@ faf-api:
   jwt:
     secretKeyPath: ${JWT_PRIVATE_KEY_PATH:test-pki-private.key}
     publicKeyPath: ${JWT_PUBLIC_KEY_PATH:test-pki-public.key}
+    hydraJwksUrl: ${JWT_HYDRA_JWKS_URL:https://hydra.test.faforever.com/.well-known/jwks.json}
```
bukajsytlos, line 6: is hydra somehow relevant in this context? isn't it implementation detail?

Brutus5000: It's not necessarily Hydra that is relevant here, but it is the OpenID Connect token that we customized (manual extension for our role system). So it's not just a default Jwks url, it must point to a faf hydra instance.

src/main/java/com/faforever/api/security/FafMultiTokenStore.java
Brutus5000: I'm not exactly sure which exceptions are thrown, I know about AccessDeniedException though

Brutus5000: Browsed through the code, there is a shitload of potential Exceptions waiting to be thrown. I'll just leave it as is. It does no harm.

src/main/resources/config/application-dev.yml
Brutus5000: Slightly renamed it to emphasize it's faf hydra not any hydra

bukajsytlos: I still don't understand, if we change hydra for octa or something else, wouldn't we need just to change the value and not the key?

PR #444, opened 2021-03-14
Feature/#443 add map pool assignment

### LINKED ISSUE #443: Allow reading and writing of map params in db v109

Sheikah45:
Closes #443

### DIFF
--- a/src/inttest/java/com/faforever/api/data/MapPoolElideTest.java
+++ b/src/inttest/java/com/faforever/api/data/MapPoolElideTest.java
@@ -16,29 +16,30 @@ import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.
 @Sql(executionPhase = ExecutionPhase.BEFORE_TEST_METHOD, scripts = "classpath:sql/prepDefaultData.sql")
 @Sql(executionPhase = ExecutionPhase.BEFORE_TEST_METHOD, scripts = "classpath:sql/prepMapData.sql")
 public class MapPoolElideTest extends AbstractIntegrationTest {
-  private static final String NEW_LADDER_MAP_BODY = "{" +
-    "\"data\": {" +
-    "  \"type\": \"mapPoolAssignment\"," +
-    "  \"attributes\": {" +
-    "    \"weight\": 1," +
-    "    \"mapParams\": \"{\\\"type\\\": \\\"none\\\"}\"" +
-    "  }," +
-    "  \"relationships\": {" +
-    "    \"mapPool\": {" +
-    "      \"data\": {" +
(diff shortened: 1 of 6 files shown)


src/inttest/java/com/faforever/api/data/MapPoolElideTest.java
```diff
@@ -16,12 +16,34 @@
...
+          "type": "mapVersion",
+          "id": "1"
+        }
+      }
+    }
+  }
+}
+   */
+
+  private static final String NEW_LADDER_MAP_BODY = "{\"data\":{\"type\":\"mapPoolAssignment\",\"relationships\":{\"mapPool\":{\"data\":{\"type\":\"mapPool\",\"id\":\"1\"}},\"mapVersion\":{\"data\":{\"type\":\"mapVersion\",\"id\":\"1\"}}}}}";
```
Brutus5000: Use multiline String in the variable instead

src/main/java/com/faforever/api/data/domain/MapPoolAssignment.java
```diff
@@ -0,0 +1,65 @@
...
+@Include(rootLevel = true, type = "mapPoolAssignment")
+@CreatePermission(expression = WriteMatchmakerMapCheck.EXPRESSION)
+@UpdatePermission(expression = WriteMatchmakerMapCheck.EXPRESSION)
+@DeletePermission(expression = WriteMatchmakerMapCheck.EXPRESSION)
+public class MapPoolAssignment {
+  private int id;
+  private MapPool mapPool;
+  private MapVersion mapVersion;
+  private int weight;
+  private String mapParams;
```
Brutus5000, line 35: Do we ever care about the contents? Then maybe Map<String, Object> or something like JsonObject

Brutus5000 (changes requested):

src/main/java/com/faforever/api/data/domain/MapPoolAssignment.java
Sheikah45: Not sure what you mean here. When I try to use those types I just get that it doesn't know how to handle them.

Sheikah45:
I went ahead and made this rely on version v112 of the database so checks need to be rerun once that is released

src/main/java/com/faforever/api/data/domain/MapPoolAssignment.java
Brutus5000: Ah yes, Hibernate doesn't support this. Shame. Jackson can handle this.

Brutus5000 (approved):

PR #466, opened 2021-05-18
Fixes #465 Maps without author can be updated by anyone

### LINKED ISSUE #465: Maps without author can be updated by anyone

bukajsytlos:
Fixes #465 Maps without author can be updated by anyone

### DIFF
--- a/src/main/java/com/faforever/api/map/MapService.java
+++ b/src/main/java/com/faforever/api/map/MapService.java
@@ -45,7 +45,6 @@ import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


src/main/java/com/faforever/api/map/MapService.java
```diff
@@ -343,11 +342,9 @@ private void validateScenarioLua(MapLuaAccessor mapLua, MapNameBuilder mapNameBu
     Optional<Map> existingMapOptional = mapRepository.findOneByDisplayName(displayName);
     existingMapOptional
       .ifPresent(existingMap -> {
-        Optional.ofNullable(existingMap.getAuthor())
-          .filter(existingMapAuthor -> !Objects.equals(existingMapAuthor, author))
-          .ifPresent(existingMapAuthor -> {
-            throw ApiException.of(ErrorCode.MAP_NOT_ORIGINAL_AUTHOR, existingMap.getDisplayName());
-          });
+        final Player existingMapAuthor = existingMap.getAuthor();
+        if (existingMapAuthor == null || !existingMapAuthor.equals(author))
```
Sheikah45: Objects.equals(existingMapAuthor, author)

bukajsytlos: I don't want nulls to be equal just for sure

Brutus5000: It's also more expressively, as it indicates that existingMapAuthor can be null

Brutus5000: I think it was code style convention to always use brackets for if/else even for one liners

Brutus5000 (approved):

src/main/java/com/faforever/api/map/MapService.java
bukajsytlos: yeah not sure how this has happened :), I don't like bracketless ifs too

Sheikah45: Yeah all good I had thought there was an explicit null check on author that threw an error if null coming into the function but it isn't a big deal

PR #471, opened 2021-05-22
Deploy all files in the repository

### LINKED ISSUE #406: Auto-deployment ignores files on top level
Currently only the content in folders is being zipped and shipped. Due to that the game tab is unable to update new init.lua files

Sheikah45:
Fixes #406 

Need to test that the game launches with this deployment

### DIFF
--- a/src/main/java/com/faforever/api/deployment/LegacyFeaturedModDeploymentTask.java
+++ b/src/main/java/com/faforever/api/deployment/LegacyFeaturedModDeploymentTask.java
@@ -16,6 +16,7 @@ import lombok.SneakyThrows;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.compress.archivers.zip.ZipArchiveEntry;
 import org.apache.commons.compress.archivers.zip.ZipArchiveOutputStream;
+import org.apache.commons.io.FilenameUtils;
 import org.springframework.beans.factory.config.BeanDefinition;
 import org.springframework.context.annotation.Scope;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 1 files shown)


Sheikah45:
Tested on the test server and the game launched properly for me with the repo init file

Sheikah45:
This should be ready now

Sheikah45:
Bump

src/main/java/com/faforever/api/deployment/LegacyFeaturedModDeploymentTask.java
```diff
@@ -198,12 +199,20 @@ private void updateStatus(String message) {
...
+          if (Files.isDirectory(path) && !path.getFileName().toString().startsWith(".")) {
+            return packDirectory(path, version, targetFolder, fileIds);
+          } else if (Files.isRegularFile(path)) {
+            return packFile(path, version, targetFolder, fileIds);
+          } else{
+            return Optional.empty();
+          }
+        })
         .filter(Optional::isPresent)
         .map(Optional::get)
```
Brutus5000: instead of map you can flatMap other streams.

packDirectory returns an Optional, so we can call .stream()
packFile also returns an Optional, 
Optional.empty can be replaced with stream.empty

This gets rid of 
        .filter(Optional::isPresent)
        .map(Optional::get)

Brutus5000 (changes requested):

src/main/java/com/faforever/api/deployment/LegacyFeaturedModDeploymentTask.java
Sheikah45: Good point.

Brutus5000 (approved):

PR #476, opened 2021-05-28
Update commons

Sheikah45:
tied to https://github.com/FAForever/faf-java-commons/pull/48


Brutus5000 (changes requested):
The commit id has changed after my merges.

PR #481, opened 2021-06-07
Code style improvements

Brutus5000:


src/main/java/com/faforever/api/deployment/LegacyFeaturedModDeploymentTask.java
```diff
@@ -217,7 +218,7 @@ private void updateStatus(String message) {
         .map(path -> packDirectory(path, version, targetFolder, fileIds))
         .filter(Optional::isPresent)
         .map(Optional::get)
-        .collect(Collectors.toList());
+        .collect(Collectors.toCollection(ArrayList::new));
```
Sheikah45, line 227: Why use the expanded syntax here? Is it just to make it clearer?

Brutus5000: Sort of. Actually it follows the API to prevent it from breaking in the future. The javadoc of `Collectors.toList()` says there is no guarantee on the type of the list being returned and if you need a special type of list (in our case: a mutable one) you need to specify it like this.

Sheikah45: Ah gotcha that makes sense. It didn't occur to me they could change the mutability of it.

Brutus5000: Unfortunately List.of() and similar operators do provide ImmutableList without the type system reflecting it. The collection interfaces of Java are an unfixable mess when it comes to mutability. Instead of the type system having your back, you'll just run into runtime errors if you get handed an immutable list. But for compatibility reasons there will never be an official solution.

src/main/java/com/faforever/api/deployment/GitHubDeploymentService.java
```diff
@@ -40,20 +39,20 @@ public GitHubDeploymentService(ApplicationContext applicationContext, FafApiProp
   void createDeploymentIfEligible(Push push) {
     String ref = push.getRef();
     Optional<FeaturedMod> optional = featuredModService.findByGitUrlAndGitBranch(
-      push.getRepository().gitHttpTransportUrl(),
+      push.getRepository().getHttpTransportUrl(),
```
Sheikah45, line 42: haha that was a nice typo in the library

Sheikah45 (approved):
Looks good to me. Although I do find it odd how with records they decided to change the getter syntax. Will take some getting used to.

bukajsytlos (approved):
lgtm

PR #483, opened 2021-06-11
Hack in support for OpenID scopes

Brutus5000:
OpenID token have a different structure than the Spring ones. This workaround makes the API support both.

It's really ugly but the whole OAuth2 code of the API is deprecated anyway.


bukajsytlos (commented):
lgtm

PR #484, opened 2021-06-11
Enable voting for very old accounts with few valid games

Brutus5000:
(e.g. single player fanatics).
Lex Uveso

### DIFF
--- a/src/main/java/com/faforever/api/voting/VotingService.java
+++ b/src/main/java/com/faforever/api/voting/VotingService.java
@@ -25,6 +25,7 @@ import java.util.Set;
 @Service
 @RequiredArgsConstructor
 public class VotingService {
+  public static final int ACCOUNT_AGE_YEARS_VOTE_QUALIFIED = 3;
   private final VoteRepository voteRepository;
   private final VotingSubjectRepository votingSubjectRepository;
   private final GamePlayerStatsRepository gamePlayerStatsRepository;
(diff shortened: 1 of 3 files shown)


src/main/java/com/faforever/api/voting/VotingService.java
```diff
@@ -100,7 +101,9 @@ public void saveVote(Vote vote, Player player) {
       errors.add(new Error(ErrorCode.VOTE_ALREADY_ENDED, subject.getEndOfVoteTime()));
     }
 
-    if (gamesPlayed < subject.getMinGamesToVote()) {
+    boolean accountQualifiedBySteamAndAge = player.getSteamId() != null && OffsetDateTime.now().minusYears(ACCOUNT_AGE_YEARS_VOTE_QUALIFIED).isAfter(player.getCreateTime());
+
+    if (!accountQualifiedBySteamAndAge && gamesPlayed < subject.getMinGamesToVote()) {
       errors.add(new Error(ErrorCode.NOT_ENOUGH_GAMES, gamesPlayed, subject.getMinGamesToVote()));
```
bukajsytlos, line 107: Shouldn't we change error message too?

bukajsytlos (commented):
No tests? :(

src/main/java/com/faforever/api/voting/VotingService.java
Brutus5000: No, **if** the error occurs its still correct. It just occurs in less cases.

bukajsytlos: wah, you are right

bukajsytlos (approved):

PR #493, opened 2021-07-30
Fix no scope detected for hydra tokens

Sheikah45:
Split classic and hydra jwtAccessTokenConverters so they can each have their own scope attribute


Sheikah45:
To explain the change a little more.

The classic tokens have their scope under the claim `scope` while the hydra tokens have the scope under the claim `scp`. So while we have the custom authentication converter which extracts the roles under the claim`ext` which is present in both tokens we do not get the scopes if we use the same converter for both.

Luckily the default converter can easily change what claim has the scope through the scope attribute property. So a converter for each type of token is created to be used with the appropriate store.

Sheikah45:
This issue happens because the hasScope check does not check the authorities which is populated by the custom token converter. Rather it checks a separate scopes property which is populated based on the scope claim.

Brutus5000 (approved):

PR #498, opened 2021-08-22
Use generic in AbstractEntity to allow proper setter chaining

Sheikah45:


bukajsytlos:
Will this work? I think we need field level annotations to be able to use custom setters. And I think lombok has something like this. (@Accessors)

Sheikah45:
The chaining is already set in the lombok.config. The issue is just that it returns an AbstractEntity rather than the class it is used on

Sheikah45:
It also works fine when I run it locally as I think elide only looks for the naming conventions when set at the property level

bukajsytlos (approved):

Brutus5000 (approved):
I love recursions

PR #500, opened 2021-08-30
Feature/gog linking

Geosearchef:
API implementation is done and integration tested, still need to write unit tests.
Website/frontend is functional, need to do a bit of reworking on that though.

website pull: https://github.com/FAForever/website/pull/291/

### DIFF
--- a/src/main/java/com/faforever/api/user/GogService.java
+++ b/src/main/java/com/faforever/api/user/GogService.java
@@ -1,0 +1,17 @@
+package com.faforever.api.user;
+
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Service;
+
+import java.util.regex.Pattern;
+
+@Service
+@Slf4j
+@RequiredArgsConstructor
+public class GogService {
+
+  private static final Pattern GOG_USERNAME_PATTERN = Pattern.compile("[\w]+");
+
+
+}
--- a/src/main/java/com/faforever/api/user/GogService.java
+++ b/src/main/java/com/faforever/api/user/GogService.java
@@ -4,9 +4,8 @@ import com.faforever.api.config.FafApiProperties;
 import com.faforever.api.data.domain.User;
 import com.faforever.api.error.ApiException;
 import com.faforever.api.error.ErrorCode;
-import com.google.gson.Gson;
-import lombok.AllArgsConstructor;
-import lombok.Data;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.jsoup.Jsoup;
@@ -17,29 +16,24 @@ import org.springframework.stereotype.Service;
 import org.springframework.web.client.HttpClientErrorException;
 import org.springframework.web.client.RestTemplate;
 
-import java.io.Serializable;
-import java.util.ArrayList;
 import java.util.List;
-import java.util.Map;
 import java.util.Objects;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-import java.util.stream.Collectors;
 
 @Service
 @Slf4j
 @RequiredArgsConstructor
 public class GogService {
-  // TODO: what characters to allow?
-  private static final Pattern GOG_USERNAME_PATTERN = Pattern.compile("[a-z\\d-_!?.]+"); // gets lower cased by controller, gog doesn't care about capitalization, this causes db entries to be unique
+  private static final Pattern GOG_USERNAME_PATTERN = Pattern.compile("[a-zA-Z\\d-_!?.]+");
   private static final Pattern PROFILE_USER_STATUS_PATTERN = Pattern.compile("window\\.profilesData\\.profileUserPreferences\\s=\\s\\{\"bio\":\"(.*?)\"");
   private static final String GOG_FA_GAME_ID = "1444785261";
 
   private final FafApiProperties properties;
-  private final Gson gson = new Gson();
+  private final RestTemplate restTemplate;
 
   boolean verifyGogUsername(String username) {
-    return username.length() <= 100 && username.length() >= 3 && GOG_USERNAME_PATTERN.matcher(username).matches();
+    return username.length() <= 30 && username.length() >= 3 && GOG_USERNAME_PATTERN.matcher(username).matches();
   }
 
   boolean verifyProfileToken(String gogUsername, User user, String targetToken) {
@@ -50,12 +44,12 @@ public class GogService {
 
   private String getProfileStatus(String gogUsername) {
     String profilePageUrl = String.format(properties.getGog().getProfilePageUrl(), gogUsername);
-    String profilePageHtml = null;
+    String profilePageHtml;
 
     try {
-      profilePageHtml = new RestTemplate().getForObject(profilePageUrl, String.class);
+      profilePageHtml = restTemplate.getForObject(profilePageUrl, String.class);
     } catch (HttpClientErrorException e) {
-      if (e.getRawStatusCode() == HttpStatus.NOT_FOUND.value()) {
+      if (e.getStatusCode() == HttpStatus.NOT_FOUND) {
         throw ApiException.of(ErrorCode.GOG_LINK_PROFILE_NOT_PUBLIC);
       } else {
         log.error("Couldn't retrieve gog profile page for {}", profilePageUrl);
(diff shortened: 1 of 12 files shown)


src/main/java/com/faforever/api/user/GogService.java
```diff
@@ -0,0 +1,177 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+@Service
+@Slf4j
+@RequiredArgsConstructor
+public class GogService {
+  // TODO: what characters to allow?
+  private static final Pattern GOG_USERNAME_PATTERN = Pattern.compile("[a-z\\d-_!?.]+"); // gets lower cased by controller, gog doesn't care about capitalization, this causes db entries to be unique
```
Geosearchef: What characters should we allow for GOG usernames? I'd prefer not to just take any string.

src/main/java/com/faforever/api/user/UserService.java
```diff
@@ -423,6 +426,40 @@ public CallbackResult linkToSteam(String token, String steamId) {
     return new CallbackResult(callbackUrl, errors);
   }
 
+  public String buildGogToken(User user) {
+    return String.format(properties.getGog().getTokenFormat(), user.getId());
+  }
+
+  public void linkGogAccount(String gogUsername, User user) { // TODO: rate limit?
```
Geosearchef: Any ideas on rate limiting so we don't send lots of external requests?

src/main/java/com/faforever/api/user/GogService.java
```diff
@@ -0,0 +1,177 @@
...
+
+    return allGames.stream().anyMatch(game -> Objects.equals(game.game.id, GOG_FA_GAME_ID));
+  }
+
+  private List<GogGamesListPage.GogGamesListEntry> getAllGames(String gogUsername) {
+    GogGamesListPage firstPage = getGamesListPage(gogUsername, 1);
+
+    List<GogGamesListPage> pages = new ArrayList<>();
+    pages.add(firstPage);
+    for (int i = 2; i <= firstPage.getPages(); i++) {
```
bukajsytlos: wouldn't it be better to search for faf on each page and break, instead of loading all games?

src/main/java/com/faforever/api/user/GogService.java
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    return pages.stream().flatMap(page -> page.get_embedded().getItems().stream()).collect(Collectors.toList());
+  }
+
+  private GogGamesListPage getGamesListPage(String gogUsername, int page) {
+    String gamesListUrl = String.format(properties.getGog().getGamesListUrl(), gogUsername, page);
+
+    try {
+      String json = new RestTemplate().getForObject(gamesListUrl, String.class);
```
bukajsytlos: RestTemplate is thread-safe so it can be extracted as field. even better would be to inject it

src/main/java/com/faforever/api/user/GogService.java
```diff
@@ -0,0 +1,177 @@
...
+      if (json == null || json.isBlank()) {
+        throw ApiException.of(ErrorCode.GOG_LINK_GAMES_NOT_PUBLIC);
+      }
+      return gson.fromJson(json, GogGamesListPage.class);
```
bukajsytlos: RestTemplate should be able to directly deserialize to object instead of string

src/main/java/com/faforever/api/user/UserService.java
```diff
@@ -423,6 +426,40 @@ public CallbackResult linkToSteam(String token, String steamId) {
...
+    if(! gogService.verifyProfileToken(gogUsername, user, buildGogToken(user))) {
+      throw ApiException.of(ErrorCode.GOG_LINK_PROFILE_TOKEN_NOT_SET);
+    }
+
+    if(! gogService.verifyGameOwnership(gogUsername)) {
+      throw ApiException.of(ErrorCode.GOG_LINK_NO_FA_GAME);
+    }
+
+    Optional<User> userWithSameId = userRepository.findOneByGogId(gogUsername);
+    if(userWithSameId.isPresent()) {
```
bukajsytlos: userWithSameId.ifPresent(user -> throw...)

src/main/java/com/faforever/api/user/UserService.java
```diff
@@ -423,6 +426,40 @@ public CallbackResult linkToSteam(String token, String steamId) {
...
+      throw ApiException.of(ErrorCode.GOG_ID_ALREADY_LINKED, userWithSameId.get().getLogin());
+    }
+
+    if(user.getGogId() != null) {
```
bukajsytlos: this will never happen as this is covered by previous case, right?

bukajsytlos: this could be the first check so that no calls are made to gog

bukajsytlos (changes requested):

src/main/java/com/faforever/api/user/GogService.java
Brutus5000: I tried to figure it out with fake registrations but they only tell you that there are invalid special characters, not which ones.

src/main/java/com/faforever/api/user/GogService.java
```diff
@@ -0,0 +1,177 @@
...
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+@Service
+@Slf4j
+@RequiredArgsConstructor
+public class GogService {
+  // TODO: what characters to allow?
+  private static final Pattern GOG_USERNAME_PATTERN = Pattern.compile("[a-z\\d-_!?.]+"); // gets lower cased by controller, gog doesn't care about capitalization, this causes db entries to be unique
+  private static final Pattern PROFILE_USER_STATUS_PATTERN = Pattern.compile("window\\.profilesData\\.profileUserPreferences\\s=\\s\\{\"bio\":\"(.*?)\"");
```
Brutus5000, line 30: Use the multiline string (`"""`) and avoid some of the escaping

src/main/java/com/faforever/api/user/GogService.java
```diff
@@ -0,0 +1,177 @@
...
+  private static final String GOG_FA_GAME_ID = "1444785261";
+
+  private final FafApiProperties properties;
+  private final Gson gson = new Gson();
+
+  boolean verifyGogUsername(String username) {
+    return username.length() <= 100 && username.length() >= 3 && GOG_USERNAME_PATTERN.matcher(username).matches();
```
Brutus5000: from my tests length must be <30 for sure.

src/main/java/com/faforever/api/user/GogService.java
```diff
@@ -0,0 +1,177 @@
...
+
+    return profileStatus.length() < 100 && Objects.equals(targetToken, profileStatus.trim());
+  }
+
+  private String getProfileStatus(String gogUsername) {
+    String profilePageUrl = String.format(properties.getGog().getProfilePageUrl(), gogUsername);
+    String profilePageHtml = null;
+
+    try {
+      profilePageHtml = new RestTemplate().getForObject(profilePageUrl, String.class);
```
Brutus5000: please get the RestTemplate via dependency injection

src/main/resources/config/application.yml
```diff
@@ -27,6 +27,10 @@ faf-api:
     base-url: ${NODEBB_BASE_URL:false}
     admin-user-id: ${NODEBB_ADMIN_USERID:1}
     master-token: ${NODEBB_MASTER_TOKEN:false}
+  gog:
+    token-format: ${GOG_TOKEN_FORMAT:FAF_%s}
```
Brutus5000: I'd prefer something more "enclosing" such as `{{FAF:<id>}}` or `[FAF]<id>[/FAF]` (from my test the `/` would need escaping though)

src/main/resources/config/application.yml
```diff
@@ -27,6 +27,10 @@ faf-api:
...
+    profile-page-url: ${GOG_PROFILE_URL:https://www.gog.com/u/%s}
+    games-list-url: ${GOG_GAMES_LIST_URL:https://www.gog.com/u/%s/games/stats?sort=recent_playtime&order=desc&page=%d}
```
Brutus5000, line 33: you can remove sort and order params to make it more comprehensible (I tested it, it works)

src/main/java/com/faforever/api/user/UserService.java
Brutus5000: This method should be @Transactional.

To your question: No, I would not rate limit.

src/main/java/com/faforever/api/user/UserService.java
```diff
@@ -87,7 +89,7 @@ public UserService(EmailService emailService,
                      AnopeUserRepository anopeUserRepository,
                      FafTokenService fafTokenService,
                      SteamService steamService,
-                     GlobalRatingRepository globalRatingRepository,
+                     GogService gogService, GlobalRatingRepository globalRatingRepository,
```
Brutus5000: line break

src/main/java/com/faforever/api/user/GogService.java
```diff
@@ -0,0 +1,177 @@
...
+
+  @Data
+  @AllArgsConstructor
+  private static class GogGamesListPage implements Serializable {
+    private int page;
+    private int limit;
+    private int pages;
+    private int total;
+    private GogGamesListLinkList _links;
+    private GogGamesListEmbeddedList _embedded;
```
Brutus5000: use proper java variable names and use @JsonProperty to indicate the json field name

src/main/java/com/faforever/api/user/GogService.java
```diff
@@ -0,0 +1,177 @@
...
+        throw ApiException.of(ErrorCode.GOG_LINK_USER_NOT_FOUND);
+      } else {
+        throw e;
+      }
+    }
+  }
+
+  @Data
+  @AllArgsConstructor
+  private static class GogGamesListPage implements Serializable {
```
Brutus5000: I would remove all attributes that we do not care about. If this causes troubles, we propably need `@JsonIgnoreProperties(ignoreUnknown = true)` on class level

src/main/java/com/faforever/api/user/GogService.java
Brutus5000: same as above

src/main/java/com/faforever/api/user/GogService.java
```diff
@@ -0,0 +1,177 @@
...
+  private List<GogGamesListPage.GogGamesListEntry> getAllGames(String gogUsername) {
+    GogGamesListPage firstPage = getGamesListPage(gogUsername, 1);
+
+    List<GogGamesListPage> pages = new ArrayList<>();
+    pages.add(firstPage);
+    for (int i = 2; i <= firstPage.getPages(); i++) {
+      pages.add(getGamesListPage(gogUsername, i));
+    }
+
+    return pages.stream().flatMap(page -> page.get_embedded().getItems().stream()).collect(Collectors.toList());
```
Brutus5000: This is a bit ugly because you even query all pages if your desired result is on the first page.
But I don't have a class Stream solution for Java 8 at hand.

Brutus5000 (changes requested):

src/main/resources/config/application.yml
bukajsytlos: wouldn't it make finding FA more probable? or is it sorted that way by default?

src/main/java/com/faforever/api/user/UserService.java
Geosearchef: ah, right, that was it, I remembered something, but wasn't sure what it was. Also, sry @transactional

src/main/java/com/faforever/api/user/GogService.java
Geosearchef: It can, but their API will return an empty string (not null) when the profile is not public.

src/main/java/com/faforever/api/user/UserService.java
Geosearchef: will catch the exception that I'm trying to throw, I had that before

bukajsytlos: hm how can it catch? doesn't make sense to me

src/main/java/com/faforever/api/user/GogService.java
Brutus5000: I take it bad. The implementation of JDK is horrible here and doesn't work at all in this case.

src/main/java/com/faforever/api/user/GogService.java
Brutus5000: I take this back. The Java implementation version doesn't work here at all (single line and ending with quotes just breaks)

bukajsytlos (dismissed):
lgtm

src/main/java/com/faforever/api/user/GogService.java
Brutus5000: No, it will raise a 403 instead and ignore the empty string.

PR #502, opened 2021-09-29
Provide access to placementGames in leagueSeason

BlackYps:
The upcoming release of the league service will have the number of placement games in the season so the client can read it and display it to the player.


bukajsytlos (dismissed):

BlackYps:
Seems like some tests fail because of the new column. I'm not familiar with these kind of tests. How do I fix them?

Brutus5000:
You need to update the league service version here: https://github.com/FAForever/faf-java-api/blob/3fc1e24ae66d6d8e101c9d4f992cd72442044087/src/inttest/java/com/faforever/api/league/LeagueAbstractIntegrationTest.java#L16

bukajsytlos (dismissed):

Brutus5000:
@BlackYps Please fix conflict and we're ready to go

PR #509, opened 2021-11-01
Provide image urls for division images

BlackYps:
The next version of the league service will provide the urls of the folders containing the division images, so the client doesn't have to store it. This provides flexibility if we ever want to change the url.


Brutus5000 (commented):
I think this needs an update (key -> url)

BlackYps:
coincidentally I'm working on it right now, but I got distracted by a certain ci problem with the league service :D

BlackYps:
So here is my first try on computing the final url for the subdivisions. Some questions:
I noticed that other enrichers pull the basae url from the FafApiProperties. How are they set? Seems to be like basic configuration not stored in the db.
I think it doesn't apply in my case, because the base urls are actually stored in a db table. Do I still need some annotation on the enhance method?
Also do we want to expose the base urls on the league object? It is not really necessary, but may be nice to have.

Brutus5000:
> I noticed that other enrichers pull the basae url from the FafApiProperties. How are they set? Seems to be like basic configuration not stored in the db.

No they are part of the application propertiers classes below FafApiProperties (either application.yaml or read from env variables).

> I think it doesn't apply in my case, because the base urls are actually stored in a db table. Do I still need some annotation on the enhance method?

No in that case you don't need it.

BlackYps:
I talked with Sheikah and we agreed that it is not really desireable to have the base urls exposed on the league endpoint. Is there a way how I can access them for the enricher without having them public?

BlackYps:
@Brutus5000

bukajsytlos:
lgtm, but I don't like the concept of storing base paths in DB, when it is more runtime configuration thing

bukajsytlos (approved):
lgtm, if concept is ok

Brutus5000 (approved):

PR #510, opened 2021-12-04
Provide season number

BlackYps:
During review of the client ui code we noticed that having the number of the season explicitly stored makes internationalization a lot easier.


BlackYps:
I don't understand why codacy believes to have found a syntax error...

bukajsytlos (approved):

Brutus5000 (approved):

PR #536, opened 2022-01-03
Add population of the service links table

Sheikah45:
@Brutus5000 let me know if this is how you thought of it. 

Dependent on https://github.com/FAForever/db/pull/294


Brutus5000 (dismissed):

Brutus5000 (commented):
We should not yet remove steamid from the entity and rather mark it as deprecated (and/or populate it with value deprecated) in order to avoid breaking compatibility with other apps (moderator client)

Sheikah45:
Do you mean just for visibility sake?
 Because it wouldn't break the mod client the Id would just be null in the user management view

PR #546, opened 2022-01-16
Put all tests into testcontainers

Brutus5000:
Open point: Make it optional to run integration tests against real databases for performance reason. (The overhead of testcontainer db migrations for a single test is significant)


bukajsytlos (dismissed):
lgtm :D:D

src/inttest/java/com/faforever/api/config/LeagueDbTestContainers.java
```diff
@@ -0,0 +1,58 @@
...
+
+  static {
+    leagueServiceDBContainer
+      .withNetwork(sharedNetwork)
+      .withNetworkAliases("faf-league-db")
+      .withUsername("faf-league-service")
+      .withPassword("banana")
+      .withDatabaseName("faf-league")
+      .withReuse(true)
+      .start();
```
Brutus5000, line 51: Isn't this always started because it's a static block?

bukajsytlos: I thought so too, but apparently none of these are met https://docs.oracle.com/javase/specs/jls/se7/html/jls-12.html#jls-12.4.1 and it is not invoked

PR #561, opened 2022-02-12
Beautiful emails (fixes #560)

### LINKED ISSUE #560: Beautiful html emails
We give up on Mautic (https://github.com/FAForever/faf-stack/issues/199).

Instead I stumbled onto this beautiful project: https://mjml.io/
This should allow making beautiful emails possible even for design cretins just like me.

- [ ] Account activation
- [ ] Password reset
- [ ] (new) Welcome email after activation

Brutus5000:
Add FAF styled emails in MailJet markup language
Load emails at runtime from html templates

### DIFF
--- a/src/main/java/com/faforever/api/email/MailBodyBuilder.java
+++ b/src/main/java/com/faforever/api/email/MailBodyBuilder.java
@@ -1,0 +1,144 @@
+package com.faforever.api.email;
+
+import com.faforever.api.config.FafApiProperties;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.InitializingBean;
+import org.springframework.stereotype.Component;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static java.text.MessageFormat.format;
+
+@Component
+@RequiredArgsConstructor
+@Slf4j
+public class MailBodyBuilder implements InitializingBean {
+
+  @Getter
+  public enum Template {
+    ACCOUNT_ACTIVATION("username", "activationUrl"),
+    WELCOME_TO_FAF("username"),
+    PASSWORD_RESET("username", "passwordResetUrl");
+
+    private final Set<String> variables;
+
+    Template(String... variables) {
+      this.variables = Set.of(variables);
+    }
+  }
+
+  private final FafApiProperties properties;
+
+  private Path getTemplateFilePath(Template template) {
+    final var path = switch (template) {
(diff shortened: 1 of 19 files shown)


src/main/java/com/faforever/api/email/MailBodyBuilder.java
```diff
@@ -0,0 +1,145 @@
...
+  @Getter
+  public enum Template {
+    ACCOUNT_ACTIVATION("username", "activationUrl"),
+    WELCOME_TO_FAF("username"),
+    PASSWORD_RESET("username", "passwordResetUrl");
+
+    private final Set<String> variables;
+
+    Template(String... variables) {
+      this.variables = Arrays.stream(variables).collect(Collectors.toSet());
```
bukajsytlos:
```suggestion
      this.variables = Set.of(variables);
```

src/main/java/com/faforever/api/email/MailBodyBuilder.java
```diff
@@ -0,0 +1,145 @@
...
+
+import static java.text.MessageFormat.format;
+
+@Component
+@RequiredArgsConstructor
+@Slf4j
+public class MailBodyBuilder implements InitializingBean {
+
+  @Getter
+  public enum Template {
```
bukajsytlos, line 25: I am just thinking about using abstract class as a template with constructor with a path to template or direct string of template and method format(Map<> variables), and have each mail type as subclass. This class would do all the validation. But I am too tired to sort out all the details :)

bukajsytlos (dismissed):

src/main/java/com/faforever/api/email/MailBodyBuilder.java
Brutus5000: Yeah. Sounds like a use case for sealed classes.

Brutus5000: So after looking at it I see not better solution

Moving away from enum to sealed classes as typed data carriers (carrying the variables) we lose the ability to iterate over the existing list of classes without reflection + we'd need to instantiate them to have template validation on startup.
Having each class implement a format method leads to code duplication unless you abstract it but then you end up with the same method as of now.

The only differences in the templates are: 
* different configuration in the properties (which imho shouldn't be coded into a POJO - doesn't make sense you either need to inject the properties or pass them in the method)
* different variables: but we abstracted this nicely with the public methods of the MailBodyBuilder methods

Using more classes felt like doing OOP for the sake of OOP without adding benefits. I'm open to improvement PRs though if you find a better solution.

src/main/java/com/faforever/api/email/MailBodyBuilder.java
Brutus5000: Si. I always get confused with vararg methods.

bukajsytlos (approved):

PR #573, opened 2022-03-12
Remove double negative in modService

Sheikah45:
This really bothered me when I saw it XD


Brutus5000 (commented):
Your change is not semantically equivalent.

true && false
is not equivalent to
!(true && true)

So does this fix a bug or not?

Sheikah45:
Bah you are right forgot it depends on if the mod exists at this point and not just if I am the uploader

PR #633, opened 2022-07-24
Receive AchievementUpdates from RabbitMQ

Brutus5000:
Using RabbitMQ for unidirectional messages has multiple advantages:
- No messages get lost if the API is not available
- Error messages get parked in a dead letter queue, so they can be reprocessed after fixing
- No handling of server-to-server OAuth handling (permissions to RabbitMQ are implicit)


src/main/java/com/faforever/api/config/RabbitConfiguration.java
```diff
@@ -0,0 +1,118 @@
...
+    factory.setMessageConverter(new Jackson2JsonMessageConverter());
+
+    return factory;
+  }
+
+  /**
+   * Declare all rabbitmq objects
+   */
+  @Bean
+  public Declarables declarables() {
```
bukajsytlos, line 76: does this need to be public when other `@Bean` are not?

src/main/java/com/faforever/api/config/RabbitConfiguration.java
```diff
@@ -0,0 +1,118 @@
...
+      exchanges,
+      queues,
+      bindWithDlq(fafLobbyExchange, deadLetterExchange, achievementQueue, QUEUE_ACHIEVEMENT_ROUTING_KEY)
+    );
+  }
+
+  /**
+   * Creates a binding for a queue along with a matching DLQ and DLQ binding
+   */
+  public List<Declarable> bindWithDlq(Exchange fromExchange, Exchange dlqExchange, Queue queue, String routingKey) {
```
bukajsytlos: could this be private?

src/main/java/com/faforever/api/config/RabbitConfiguration.java
```diff
@@ -0,0 +1,118 @@
...
+      .bind(dlq)
+      .to(dlqExchange)
+      .with(routingKey)
+      .and(Map.of("x-dead-letter-exchange", dlqExchange.getName()));
+
+    return List.of(queueBinding, dlq, dlqBinding);
+  }
+
+  @SafeVarargs
+  public final Declarables merge(List<Declarable>... declarables) {
```
bukajsytlos: this also (private)

bukajsytlos (dismissed):

src/main/java/com/faforever/api/config/RabbitConfiguration.java
Brutus5000: correct. also final was stupid :D

src/main/java/com/faforever/api/config/RabbitConfiguration.java
Brutus5000: yes

src/main/java/com/faforever/api/config/RabbitConfiguration.java
Brutus5000: I checked the other configs and we went for all beans are public. so I made the other ones public instead

bukajsytlos (dismissed):
lgtm

Askaholic:
So what are the routing keys for achievements and events? And how do we get the confirmation responses?

PR #637, opened 2022-08-06
Add HMAC verification to patch files for cloudflare

Sheikah45:
This adds verification of the patch files using cloudflares hmac authentication token as specified here https://support.cloudflare.com/hc/en-us/articles/115001376488-Configuring-Token-Authentication

This will allow us to limit access to our patch files only to those who have proven game ownership

### DIFF
--- a/src/inttest/java/com/faforever/api/featuredmods/FeaturedModsControllerTest.java
+++ b/src/inttest/java/com/faforever/api/featuredmods/FeaturedModsControllerTest.java
@@ -1,0 +1,26 @@
+package com.faforever.api.featuredmods;
+
+import com.faforever.api.AbstractIntegrationTest;
+import org.junit.jupiter.api.Test;
+import org.springframework.test.context.jdbc.Sql;
+
+import static org.hamcrest.Matchers.hasSize;
+import static org.hamcrest.Matchers.is;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.jsonPath;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+@Sql(executionPhase = Sql.ExecutionPhase.BEFORE_TEST_METHOD, scripts = "classpath:sql/truncateTables.sql")
+@Sql(executionPhase = Sql.ExecutionPhase.BEFORE_TEST_METHOD, scripts = "classpath:sql/prepDefaultData.sql")
+@Sql(executionPhase = Sql.ExecutionPhase.BEFORE_TEST_METHOD, scripts = "classpath:sql/prepFeaturedMods.sql")
+public class FeaturedModsControllerTest extends AbstractIntegrationTest {
+
+  @Test
+  public void featuredModFileUrlCorrect() throws Exception {
+    mockMvc.perform(get("/featuredMods/0/files/latest")
+        .with(getOAuthTokenWithActiveUser(NO_SCOPE, NO_AUTHORITIES)))
+      .andExpect(status().isOk())
+      .andExpect(jsonPath("$.data", hasSize(1)))
+      .andExpect(jsonPath("$.data[0].attributes.url", is("USER")));
+  }
+}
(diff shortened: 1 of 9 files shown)


Sheikah45:
~~Still need to fix the tests~~ Done

src/main/java/com/faforever/api/featuredmods/FeaturedModFileEnricher.java
```diff
@@ -17,10 +29,22 @@ public void init(FafApiProperties fafApiProperties) {
...
+    // https://support.cloudflare.com/hc/en-us/articles/115001376488-Configuring-Token-Authentication
+    Mac mac = Mac.getInstance(HMAC_SHA256);
+    mac.init(new SecretKeySpec(secret.getBytes(StandardCharsets.UTF_8), HMAC_SHA256));
+    byte[] macMessage = (featuredModUri.getPath() + timeStamp).getBytes(StandardCharsets.UTF_8);
+
+    String hmacEncoded = URLEncoder.encode(new String(Base64.getEncoder().encode(mac.doFinal(macMessage)), StandardCharsets.UTF_8), StandardCharsets.UTF_8);
+    String parameter = "%d-%s".formatted(timeStamp, hmacEncoded);
 
-    featuredModFile.setUrl(String.format(urlFormat, folder, featuredModFile.getOriginalFileName()));
+    featuredModFile.setUrl(UriComponentsBuilder.fromUri(featuredModUri).queryParam("verify", parameter).build().toString());
```
Brutus5000: You could just hardcode the query param in the featuredModUri similar to `faf.featured-mod.file-url-format`

src/inttest/java/com/faforever/api/featuredmods/FeaturedModsControllerTest.java
```diff
@@ -0,0 +1,26 @@
...
+
+@Sql(executionPhase = Sql.ExecutionPhase.BEFORE_TEST_METHOD, scripts = "classpath:sql/truncateTables.sql")
+@Sql(executionPhase = Sql.ExecutionPhase.BEFORE_TEST_METHOD, scripts = "classpath:sql/prepDefaultData.sql")
+@Sql(executionPhase = Sql.ExecutionPhase.BEFORE_TEST_METHOD, scripts = "classpath:sql/prepFeaturedMods.sql")
+public class FeaturedModsControllerTest extends AbstractIntegrationTest {
+
+  @Test
+  public void featuredModFileUrlCorrect() throws Exception {
+    mockMvc.perform(get("/featuredMods/0/files/latest")
+        .with(getOAuthTokenWithActiveUser(NO_SCOPE, NO_AUTHORITIES)))
```
Brutus5000: LOBBY_SCOPE instead of NO_SCOPE?

src/main/java/com/faforever/api/featuredmods/FeaturedModFileEnricher.java
Sheikah45: Unfortunately we can't because the URL path sans token is part of the token generation process

src/inttest/java/com/faforever/api/featuredmods/FeaturedModsControllerTest.java
Sheikah45: Yep still need to fix and add tests this was just a first pass

src/main/java/com/faforever/api/featuredmods/FeaturedModFileEnricher.java
Sheikah45: Unless you mean the name of the param

Brutus5000: Yes e.g. `<url>?verify=%s`

src/inttest/resources/config/application.yml
```diff
@@ -53,6 +53,9 @@ faf-api:
     max-size-bytes: 4096
     image-width: 40
     image-height: 20
+  cloudflare:
+    hmac-secret: "banana"
+    hmac-param: "verify"
```
Brutus5000, line 58: Is this actually configurable? If it is fixed by cloudflare I wouldn't make it configurable.

Sheikah45: Yes it is configurable, technically you can set it to whatever you want. They just ask for the length of the prefix before the token in bytes which for ?verify= is 8 for example. Technically we could have no prefix but I think it makes it cleaner and also allows the url to be valid even without the firewall filter

We also set the secret in the firewall filter

Brutus5000 (approved):

Brutus5000 (approved):

PR #638, opened 2022-08-11
Add Cacheable URLs for FeatureModFiles

Sheikah45:
This adds separation of the hmac token, param, and a cacheable url so that clients can send the hmac verification as a request header which allows caching of the request in cloudflare since there is no longer any query parameters

### DIFF
--- a/src/main/java/com/faforever/api/cloudflare/CloudflareService.java
+++ b/src/main/java/com/faforever/api/cloudflare/CloudflareService.java
@@ -1,7 +1,6 @@
 package com.faforever.api.cloudflare;
 
 import com.faforever.api.config.FafApiProperties;
-import lombok.RequiredArgsConstructor;
 import org.springframework.stereotype.Service;
 
 import javax.crypto.Mac;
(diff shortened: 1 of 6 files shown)


src/main/java/com/faforever/api/cloudflare/CloudflareService.java
```diff
@@ -0,0 +1,53 @@
...
+   * <a href="https://support.cloudflare.com/hc/en-us/articles/115001376488-Configuring-Token-Authentication">here</a>
+   * @param uri uri to generate the hmac token for
+   * @return string representing the hmac token formatted as {timestamp}-{hashedContent}
+   */
+  public String generateCloudFlareHmacToken(URI uri) throws NoSuchAlgorithmException, InvalidKeyException {
+    String secret = fafApiProperties.getCloudflare().getHmacSecret();
+    long timeStamp = Instant.now().getEpochSecond();
+
+    Mac mac = Mac.getInstance(HMAC_SHA256);
+    mac.init(new SecretKeySpec(secret.getBytes(StandardCharsets.UTF_8), HMAC_SHA256));
```
Brutus5000: this should be done on initializing the bean

Sheikah45: ah good point

Sheikah45:
@Brutus5000 is this good now?

Brutus5000 (approved):

PR #663, opened 2022-10-03
Allow change of login casing

### LINKED ISSUE #662: Allow user change login/username with same value with different casing
```
Is it possible to change capitalization of a nickname? (From "vonagam" to "Vonagam".)
Can't do it simply through web interface: it says that username is taken (I assume it does case insensitive check and ignores that it is me who has that login already).
```

https://discord.com/channels/197033481883222026/202928463076786176/1026210983489843250

vonagam:
Closes #662.

### DIFF
--- a/src/main/java/com/faforever/api/user/UserService.java
+++ b/src/main/java/com/faforever/api/user/UserService.java
@@ -246,7 +246,10 @@ public class UserService {
   }
 
   private void internalChangeLogin(String newLogin, User user, String ipAddress, boolean force) {
-    validateUsername(newLogin);
+    String oldLogin = user.getLogin();
+    if (!newLogin.equalsIgnoreCase(oldLogin)) {
+      validateUsername(newLogin);
+    }
 
     if (!force) {
       int minDaysBetweenChange = properties.getUser().getMinimumDaysBetweenUsernameChange();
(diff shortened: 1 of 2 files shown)


Brutus5000 (changes requested):
This is part of the validations, as such it should be in the validateUsername method. `newLogin.equalsIgnoreCase(oldLogin)` could cause an early return there before calling userRepository.existsByLogin

Also your PR should add a unit test for this case.

vonagam:
> could cause an early return there before calling userRepository.existsByLogin

That's the point, if `newLogin.equalsIgnoreCase(oldLogin)` it means that it passes username rules and that `existsByLogin` will return true and cause false error raise, so it is ok to skip `validateUsername` call here.

All other calls to `validateUsername` do not deal with situations when a user already exists (and so no false errors from `existsByLogin` there).

Added the test. Have not run it though, not a java developer, so do not have an environment setup. If you want to move this logic to `validateUsername` with Optional second argument of an existing user, that makes sense too, close this PR in that case.

Brutus5000 (approved):

Brutus5000:
Ah shame on me. I thought the CI passed but it was just Codacy. Well... we'll see if the test works on develop :D

PR #703, opened 2023-02-10
Add support for map&mod licenses

Brutus5000:
So far I didn't find the time to actually implement defining/changing the license on upload.
This should at least make it compatible with db v127

### DIFF
--- a/src/main/java/com/faforever/api/config/IgnoreOctetStreamToObjectHttpMessageConverter.java
+++ b/src/main/java/com/faforever/api/config/IgnoreOctetStreamToObjectHttpMessageConverter.java
@@ -1,0 +1,48 @@
+package com.faforever.api.config;
+
+import com.faforever.api.map.MapUploadMetadata;
+import org.springframework.http.HttpInputMessage;
+import org.springframework.http.HttpOutputMessage;
+import org.springframework.http.MediaType;
+import org.springframework.http.converter.AbstractHttpMessageConverter;
+import org.springframework.lang.Nullable;
+import org.springframework.security.core.Authentication;
+import org.springframework.util.StreamUtils;
+import org.springframework.web.multipart.MultipartFile;
+
+import java.io.IOException;
+
+/**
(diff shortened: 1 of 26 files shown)


src/test/java/com/faforever/api/map/MapServiceTest.java
```diff
@@ -348,10 +353,11 @@ void invalidScenario() {
     }
 
     @Test
-    void positiveUploadTest() throws Exception {
+    void positiveUploadTest(@Mock License licenseMock) throws Exception {
```
bukajsytlos: why mock return value? isn't creating actual test object better?

Brutus5000: Since in this PR we always set a default value, I see not much value there. It was just to make the test pass ;)

bukajsytlos:
Should the user be allowed to change license during version upload?

Brutus5000:
@bukajsytlos Only with a less strict license. Thus the helper method `isLessPermissiveThan` in the License class

src/main/java/com/faforever/api/config/IgnoreOctetStreamToObjectHttpMessageConverter.java
```diff
@@ -0,0 +1,40 @@
...
+import org.springframework.http.HttpInputMessage;
+import org.springframework.http.HttpOutputMessage;
+import org.springframework.http.MediaType;
+import org.springframework.http.converter.AbstractHttpMessageConverter;
+import org.springframework.lang.Nullable;
+import org.springframework.util.StreamUtils;
+
+import java.io.IOException;
+
+public class IgnoreOctetStreamToObjectHttpMessageConverter extends AbstractHttpMessageConverter<byte[]> {
```
Brutus5000, line 21: Please some javadoc what this does and why it's needed :)

bukajsytlos: done

PR #705, opened 2023-02-10
Prevent OOM due to caching values >1MB

Brutus5000:
@Sheikah45 I'd love to see a cross-check whether we cover all "client startup calls" correctly


Sheikah45:
What do you mean by startup calls? 

The only call the client makes on startup is to /me.

The only other critical one is the featured mods and coturn endpoints so people can start a game

Brutus5000:
Avatars and/or players or ratings after joining chat? Maps / mods that are in the current game list?

Sheikah45:
On that basis I think the ones I don't understand are the avatar assignment. As that is going to be different for everyone, it is also an infrequent call so may not even be worth caching. And then also the leaderboard rating I am not sure of either.

The map statistics isn't used by the client either I believe.

PR #719, opened 2023-03-07
Closes #652 Add replay rating to the replay search

### LINKED ISSUE #652: Add replay rating to the replay search
There is no way to search for replays which have a rating except by looking at the most highly rated games. This is not really a search though, simply a list of replays. 

I hope that if this were implemented the replay rating feature would be used more, it is not used very often at the moment.

bukajsytlos:
Closes #652 Add replay rating to the replay search


Sheikah45:
How intensive is the formula calculation? Does it perform that in SQL?

bukajsytlos:
yes it is done on DB side, but now that I think of it, we should probably make it "real" virtual column in DB so that it can be indexed

Brutus5000 (dismissed):

Brutus5000 (approved):

PR #772, opened 2023-08-18
Add gameId column to LeagueScoreJournal

BlackYps:
Requires league service 1.5.0

With this the loginId and gameId just exist as integers. They don't have a relationship to the actual object with that Id. How can this be added given that we would cross databases with this?

### DIFF
--- a/src/inttest/java/com/faforever/api/config/LeagueDbTestContainers.java
+++ b/src/inttest/java/com/faforever/api/config/LeagueDbTestContainers.java
@@ -20,7 +20,7 @@ import javax.sql.DataSource;
 @Configuration
 public class LeagueDbTestContainers {
   private static final MariaDBContainer<?> leagueServiceDBContainer = new MariaDBContainer<>("mariadb:10.6");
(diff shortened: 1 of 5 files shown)


bukajsytlos (approved):
lgtm

Brutus5000 (approved):

Brutus5000:
> They don't have a relationship to the actual object with that Id. How can this be added given that we would cross databases with this?

To our knowledge you can't and have to do 2 queries instead. I raised the question in the Elide Discord though. Although I recently discovered in MySQL you can cross join databases if you prefix the tables with the db name. Not sure if this is supported by Hibernate, then we could just drop the Multiplex store altogether...

BlackYps:
Is doing two queries still the way to go?

Brutus5000:
Currently yes, but I found a way that MySQL can do crossover joins if a user actually has permissions for both tables.
So we could actually drop the 2nd elide datastore and then we could do regular elide joins again. But time time time. :(

PR #802, opened 2023-11-24
Allow API access to Elide for services without user authentication

### LINKED ISSUE #793: Allow API access to Elide for services without user authentication
## Problem description

Due to ongoing DDoS attacks against API endpoints we protected the `/data` endpoints with `@Secured("ROLE_USER")`. However, this breaks external applications that were using the public parts of the API, such as the faf-website.

What we need is a way to authenticate services without a user context, so that these services can use the API just as before by providing _some_ kind of authentication.

## Solution ideas
There are 2 general approaches:

- Static API keys
- Oauth Access tokens for Machine2Machine authentication (hence no OpenID Connect, as there is no user component here).

OAuth access token seems the preferred approach, as along with their scopes, there is already a mechanism to limit the capabilities of access (if required). Also it is a global approach that could be used by all services wheres API key logic would need to go into each consumed service.

## OAuth approach issues
The current logic of the OAuth token parsing in the `FafAuthenticationConverter` requires the token to contain an integer subject (`sub` in the jwt) containing the FAF user id.
This would not match a machine2machine OAuth access token, as it has no user context.

We could think around setting the userId to a fallback value that does not map to a real user such as `-1`.
A cleaner solution would be to split up the `FafAuthenticationToken` into a `FafUserAuthenticationToken` and a `FafMachineAuthenticationToken`.

We could also think about setting a different role for machine (e.g. `ROLE_SERVICE` instead of `ROLE_USER` if it makes sense).

The biggest impact is probably in the permission checks of Elide where there the userId is no longer guaranteed to be available.

bukajsytlos:
Closes #793 Allow API access to Elide for services without user authentication

### DIFF
--- a/src/main/java/com/faforever/api/security/AuditService.java
+++ b/src/main/java/com/faforever/api/security/AuditService.java
@@ -17,18 +17,14 @@ public class AuditService {
 
   public void logMessage(String message) {
     final String extendedMessage = userSupplier.get()
-      .map(fafAuthenticationToken -> {
-        //move to switch pattern matching with java 21
-        if (fafAuthenticationToken instanceof FafUserAuthenticationToken fafUserAuthenticationToken) {
-          return MessageFormat.format("{0} [invoked by User ''{1}'' with id ''{2}'']",
+      .map(fafAuthenticationToken ->
+        switch (fafAuthenticationToken) {
+          case FafUserAuthenticationToken fafUserAuthenticationToken -> MessageFormat.format("{0} [invoked by User ''{1}'' with id ''{2}'']",
             message, fafUserAuthenticationToken.getUsername(), fafUserAuthenticationToken.getUserId());
-        } else if (fafAuthenticationToken instanceof FafServiceAuthenticationToken fafServiceAuthenticationToken) {
-          return MessageFormat.format("{0} [invoked by Service ''{1}'']",
+          case FafServiceAuthenticationToken fafServiceAuthenticationToken -> MessageFormat.format("{0} [invoked by Service ''{1}'']",
             message, fafServiceAuthenticationToken.getServiceName());
-        } else {
-          throw new RuntimeException();
         }
-      })
+      )
       .orElseGet(() -> MessageFormat.format("{0} [invoked by Annonymous user]", message));
     log.info(extendedMessage);
   }
--- a/src/main/java/com/faforever/api/security/AuditService.java
+++ b/src/main/java/com/faforever/api/security/AuditService.java
@@ -9,16 +9,27 @@ import java.text.MessageFormat;
 @Service
 public class AuditService {
 
-  private UserSupplier userSupplier;
+  private final UserSupplier userSupplier;
 
   public AuditService(UserSupplier userSupplier) {
     this.userSupplier = userSupplier;
   }
 
   public void logMessage(String message) {
-    final FafAuthenticationToken fafAuthenticationToken = userSupplier.get();
-    String extendedMessage = MessageFormat.format("{0} [invoked by User ''{1}'' with id ''{2}'']",
-      message, fafAuthenticationToken.getUsername(), fafAuthenticationToken.getUserId());
+    final String extendedMessage = userSupplier.get()
+      .map(fafAuthenticationToken -> {
+        //move to switch pattern matching with java 21
+        if (fafAuthenticationToken instanceof FafUserAuthenticationToken fafUserAuthenticationToken) {
+          return MessageFormat.format("{0} [invoked by User ''{1}'' with id ''{2}'']",
+            message, fafUserAuthenticationToken.getUsername(), fafUserAuthenticationToken.getUserId());
+        } else if (fafAuthenticationToken instanceof FafServiceAuthenticationToken fafServiceAuthenticationToken) {
+          return MessageFormat.format("{0} [invoked by Service ''{1}'']",
+            message, fafServiceAuthenticationToken.getServiceName());
+        } else {
+          throw new RuntimeException();
+        }
+      })
+      .orElseGet(() -> MessageFormat.format("{0} [invoked by Annonymous user]", message));
     log.info(extendedMessage);
   }
 }
--- a/src/main/java/com/faforever/api/security/FafAuthenticationConverter.java
+++ b/src/main/java/com/faforever/api/security/FafAuthenticationConverter.java
@@ -6,6 +6,7 @@ import org.springframework.security.oauth2.jwt.Jwt;
 
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 
 /**
  * Jwt converter that reads scopes + custom FAF roles from the token extension.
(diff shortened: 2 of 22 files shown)


bukajsytlos:
will try to add some tests

src/main/java/com/faforever/api/security/FafAuthenticationConverter.java
```diff
@@ -1,29 +1,35 @@
...
-    return new FafAuthenticationToken(userId, username, scopes, roles);
+    String subject = extractSubject(source);
+
+    try {
+      int userId = Integer.parseInt(subject);
+      String username = extractUsername(source);
+      return new FafUserAuthenticationToken(userId, username, scopes, roles);
+    } catch (NumberFormatException e) {
+      return new FafServiceAuthenticationToken(subject, scopes);
+    }
```
Sheikah45: Do we want to perform the type discrimination off a scope or some value in ext rather than just assuming if there is an exception that it is a service?

Maybe even just checking if the subject is a numeric string might read better and convey the intention rather than now where it looks like the Service is just a fallback instead of a distinct choice.

src/main/java/com/faforever/api/security/FafServiceAuthenticationToken.java
```diff
@@ -0,0 +1,38 @@
...
+  ) {
+    super(
+      scopes,
+      ImmutableList.<FafRole>builder()
+        .add(new FafRole("SERVICE"))
+        .build()
+    );
+    this.serviceName = serviceName;
+    // since the access token was already verified, each FafAuthenticationToken is implicitly authenticated
+    this.setAuthenticated(true);
```
Sheikah45: Is this needed since it is set in the super constructor?

src/main/java/com/faforever/api/security/FafUserAuthenticationToken.java
```diff
@@ -0,0 +1,43 @@
...
+      scopes,
+      ImmutableList.<FafRole>builder()
+        .addAll(roles)
+        .add(new FafRole("USER"))
+        .build()
+    );
+    this.userId = userId;
+    this.username = username;
+    // since the access token was already verified, each FafAuthenticationToken is implicitly authenticated
+    this.setAuthenticated(true);
```
Sheikah45: Same as above

src/main/java/com/faforever/api/security/FafServiceAuthenticationToken.java
bukajsytlos: nah, copy pasta noob

src/main/java/com/faforever/api/security/AuditService.java
```diff
@@ -9,16 +9,27 @@
...
+      .map(fafAuthenticationToken -> {
+        //move to switch pattern matching with java 21
+        if (fafAuthenticationToken instanceof FafUserAuthenticationToken fafUserAuthenticationToken) {
+          return MessageFormat.format("{0} [invoked by User ''{1}'' with id ''{2}'']",
+            message, fafUserAuthenticationToken.getUsername(), fafUserAuthenticationToken.getUserId());
+        } else if (fafAuthenticationToken instanceof FafServiceAuthenticationToken fafServiceAuthenticationToken) {
+          return MessageFormat.format("{0} [invoked by Service ''{1}'']",
+            message, fafServiceAuthenticationToken.getServiceName());
+        } else {
+          throw new RuntimeException();
```
Brutus5000: Message "Unknown authentication token type"

src/main/java/com/faforever/api/security/FafAuthenticationConverter.java
Brutus5000: No I think that check is misleading. We (user-service + hydra) are the only authority to hand out token. If we mess up I'd rather see things explode than having a fallback state where only some things work but we don't really know what is going on.

Sheikah45: I think we are agreeing that we shouldn't just fallback to assuming it is a ServiceAuthenticationToken, just have to agree on how to check it and throw an exception if the discriminator fails

Brutus5000: No was the answer to "Do we want to perform ..."  :D So yes we are agreeing

Sheikah45: Then wouldn't the answer be yes we do want to perform explicit type discrimination XD

src/main/java/com/faforever/api/security/AuditService.java
bukajsytlos: ok, but this should not happen with sealed class, right?

src/main/java/com/faforever/api/security/FafAuthenticationConverter.java
bukajsytlos: well this was my best effort with token I got from hydra with client credentials.

Brutus5000: Hmm I see. I thought we already knew the token type here.

Sheikah45: No this converter class is what produces the authentication from the jwt token that we have been given. So happens as the first step in the process.

We have no information about the token besides what is in the token.

And since both the user token and client credentials token are just JWTs,  we need some kind of discriminator to tell which kind of authentication the token is representing which I suggest we put into the ext claims when we create the token in hydra.

Brutus5000: Ok so I was checking manuals and code and stuff.
Extending m2m tokens via Webhook seems to be a feature of later Hydra versions. Not relevant.
I heavily dislike parsing the subjects value (int to non-int) to distinguish m2m vs user token.

However, the simplest and most clear solution got lost in between.
All user token have a username in `ext`. All m2m token have not. I wouldn't set it to `[undefined]` but keep it null. And if it is null it's a m2m token, otherwise a user token.

src/main/java/com/faforever/api/security/AuditService.java
Brutus5000: True. Let's do a switch then and we can drop the else case.

Sheikah45: We aren't on java 21 yet XD

src/main/java/com/faforever/api/security/FafAuthenticationConverter.java
Sheikah45: could we add a `service` claim to the `ext`? Or is that just the webhook you were talking about.

Brutus5000: Currently we can'T do anything to m2m tokens.

Sheikah45: unfortunate

src/main/java/com/faforever/api/security/AuditService.java
Brutus5000: We are on develop now :P

Brutus5000 (approved):

PR #822, opened 2024-01-13
Closes #820 Improve handling of failed steam redirect validations

### LINKED ISSUE #820: Improve handling of failed steam redirect validations
On linking to Steam and on password reset via Steam, the user gets redirected to us from Steam with a code. We then ask the Steam API if the code is valid.

If it's not valid (rare case) this unfortunately leads to an internal server error (HTTP 500).
This should cause a bad request instead (HTTP 400) and the log should contain the user id and name (as we have some request in the forum that cannot be correlated to logs).

```
2024-01-08 17:01:15.724  WARN 1 --- [f895bed0-ec2d-4666-b4d6-405e6b907ce1] [io-8010-exec-87] c.f.a.e.GlobalControllerExceptionHandler : Internal server error

java.lang.IllegalArgumentException: Steam redirect could not be validated! Original response:
ns:http://specs.openid.net/auth/2.0
is_valid:false

        at com.faforever.api.user.SteamService.validateSteamRedirect(SteamService.java:82)
        at com.faforever.api.user.UsersController.requestPasswordResetViaSteam(UsersController.java:147)
```

The way to achieve this is to not throw an IllegalArgumentException, but log a useful error and throw an ApiException.of new type STEAM_LOGIN_VALIDATION_FAILED or something.

Ivan-Shaml:
Closes #820 
Note: I have implemented `SteamService#handleInvalidOpenIdRedirect` method, whit the mindset that the validation fails because an attacker tries to exploit the endpoint.

### DIFF
--- a/src/main/java/com/faforever/api/user/SteamService.java
+++ b/src/main/java/com/faforever/api/user/SteamService.java
@@ -2,9 +2,15 @@ package com.faforever.api.user;
 
 import com.faforever.api.config.FafApiProperties;
 import com.faforever.api.config.FafApiProperties.Steam;
+import com.faforever.api.data.domain.AccountLink;
+import com.faforever.api.data.domain.LinkedServiceType;
+import com.faforever.api.error.ApiException;
+import com.faforever.api.error.ErrorCode;
+
 import lombok.RequiredArgsConstructor;
 import lombok.SneakyThrows;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.StringUtils;
 import org.json.JSONObject;
 import org.springframework.stereotype.Service;
 import org.springframework.web.client.RestTemplate;
@@ -22,6 +28,7 @@ import java.util.Map;
 @RequiredArgsConstructor
 public class SteamService {
   private final FafApiProperties properties;
+  private final AccountLinkRepository accountLinkRepository;
 
   String buildLoginUrl(String redirectUrl) {
     log.debug("Building steam login url for redirect url: {}", redirectUrl);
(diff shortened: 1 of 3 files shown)


src/main/java/com/faforever/api/user/SteamService.java
```diff
@@ -73,15 +80,46 @@ void validateSteamRedirect(HttpServletRequest request) {
...
+        LinkedServiceType.STEAM).map(AccountLink::getUser).ifPresentOrElse(u ->
+          log.warn(
+              "Steam redirect could not be validated for user with id: ''{}'' and login: ''{}''. Original OpenID response:\n {}",
+              u.getId(), u.getLogin(), openIdResponseBody),
+        () ->
+          log.warn(
+            "Steam redirect could not be validated! The steam id ''{}'' does not match any account. Original OpenID response:\n {}",
+            StringUtils.deleteWhitespace(steamId).replace("'", ""), // prevent potential log poisoning attack
+            openIdResponseBody)
+      );
```
Brutus5000: Please adjust formatting so we have on call per line

```
accountLinkRepository.findOneByServiceIdAndServiceType(...)
  .map(...)
  .ifPresentOrElse(...)
```

src/main/java/com/faforever/api/user/SteamService.java
```diff
@@ -73,15 +80,46 @@ void validateSteamRedirect(HttpServletRequest request) {
...
   }
+
+  void handleInvalidOpenIdRedirect(final HttpServletRequest request, final String openIdResponseBody) {
+    boolean containsIdentityParam = request.getParameterMap().containsKey("openid.identity");
+    final String steamId;
+
+    if (containsIdentityParam)
+    {
+       steamId = parseSteamIdFromLoginRedirect(request);
+    } else {
```
Brutus5000: The `containsIdentityParam` check could be moved into `parseSteamIdFromLoginRedirect` which would then return an Optional.
Then here we could just `.orElseThrow()`

Brutus5000 (changes requested):
Looks good, just 2 minor change requests :D

src/main/java/com/faforever/api/user/SteamService.java
Ivan-Shaml: The `com.faforever.api.user.SteamService#parseSteamIdFromLoginRedirect` method is used directly in the UsersController as well:
 - `com.faforever.api.user.UsersController#requestPasswordResetViaSteam`
 - `com.faforever.api.user.UsersController#linkToSteam`

Wouldn't it be nicer that if the loging + throwing an exception (if identityParam is not contained) get moved inside of the method, because if we make the method to return an Optional, we would need to handle that  the same way on multiple places.

Also with the current implementation of the method if the `openid.identity` parameter is not contained within the request (`request.getParameter("openid.identity")` returns `null`,  we call `.subsring` on `null`, so there is a NPE), definitely the whole method `parseSteamIdFromLoginRedirect` should be revisited anyway ...

Brutus5000 (approved):

PR #823, opened 2024-01-13
Issue-Fix-#819 append: toString methods

### LINKED ISSUE #819: Add toString() methods for regularly logged entities
- [ ] User -> print id and username
- [ ] Game -> print id and title

l3002:
Fixes #819 

## Description

Overrode toString() method for following classes:

- [x] User - prints ID and Email (wasn't able to find the username for this class therefore used email instead)
- [x] Game - prints ID and Title

### DIFF
--- a/src/main/java/com/faforever/api/data/domain/User.java
+++ b/src/main/java/com/faforever/api/data/domain/User.java
@@ -23,6 +23,6 @@ public class User extends Login {
   //Overriding toString method for efficient Logging
   @Override
   public String toString(){
-    return "User(ID = " + getId() + ", Email = " + getEmail() + ")";
+    return "User(id = " + getId() + ", email = " + getEmail() + ")";
   }
 }
--- a/src/main/java/com/faforever/api/data/domain/User.java
+++ b/src/main/java/com/faforever/api/data/domain/User.java
@@ -19,4 +19,10 @@ public class User extends Login {
   public String getPassword() {
     return password;
   }
+
+  //Overriding toString method for efficient Logging
+  @Override
+  public String toString(){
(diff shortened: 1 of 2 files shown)


src/main/java/com/faforever/api/data/domain/User.java
```diff
@@ -19,4 +19,10 @@ public class User extends Login {
   public String getPassword() {
     return password;
   }
+
+  //Overriding toString method for efficient Logging
```
Brutus5000, line 23: Is there a particular reason not to use Lombok here?

Brutus5000 (changes requested):
Please just use the variable names as is. I know I wrote title in the issue description, but I wasn't aware of the variables name.

src/main/java/com/faforever/api/data/domain/User.java
l3002: Yeah, If we use the ToString Annotation for User class then that annotation will have to call the super classes toString methods recursively as the User class inherit those properties from the classes up in the heirarchy, which leads to a messy format. Something like this User(super(super(id = 1232)), super(email = xyz@gmail.com)).

This is the reason why I used the traditional way of overriding the toString() method

Brutus5000: ok. But please make the fields names lowercase :)

l3002: @Brutus5000  I've added two new commits to this PR. Could you please check if the naming convention is alright now?

Brutus5000 (approved):

PR #918, opened 2024-12-25
fix(jackson): fix deserialization for newer versions of jackson

Ivan-Shaml:
Preparation for Spring Boot v3.4
Fixes deserialization issues with newer Jackson library versions


Brutus5000 (approved):
Not really understanding what it does, since the default according to JavaDoc `typically favors use of delegating mode`, but whatever

PR #1004, opened 2025-08-28
feat(elide): limit page size based on role

Ivan-Shaml:


Sheikah45:
As a note I think this will break the client as it currently is because there are a few places it requests the 10000 page size

Ivan-Shaml:
> As a note I think this will break the client as it currently is because there are a few places it requests the 10000 page size

Ahh, so that's why some times its so slow :stuck_out_tongue:

Jokes aside, do you think if this gets adjusted problem is solved ? :smiley: 

https://github.com/FAForever/downlords-faf-client/blob/f242ab440041fd3a6568c392ab9c878258f98c2a/src/main/java/com/faforever/client/config/ClientProperties.java#L148

*PS: For the website as well*
 * https://github.com/FAForever/website/blob/4dca50832be9bb4554068f4f21bbb7532c7971aa/src/backend/services/LeaderboardRepository.js#L18
 * https://github.com/FAForever/website/blob/4dca50832be9bb4554068f4f21bbb7532c7971aa/src/backend/services/DataRepository.js#L10

Sheikah45:
Only somewhat there will till be the issue that people who don't update clients will be broken.

And there are a few places in client code that use that max page size to assume they are getting all results. So those will have to be adjusted to be paginated or actually get all results.

src/main/java/com/faforever/api/data/util/ElidePageSizeUtil.java
```diff
@@ -0,0 +1,58 @@
...
+    final int elideMaxPageSize) {
+
+    final int rolePageSizeLimit = getRolePageSizeLimit(principal, elideDefaultPageSize,
+      elideMaxPageSize);
+    final int pageSizeInRequest = getParamsPageSize(allRequestParams).orElse(elideDefaultPageSize);
+
+    if (pageSizeInRequest > rolePageSizeLimit) {
+      throw ApiException.of(ErrorCode.QUERY_INVALID_PAGE_SIZE, pageSizeInRequest,
+        rolePageSizeLimit);
+    }
```
Brutus5000: I would not throw an error here. Instead just go for min(requested or default, allowed)

Brutus5000 (changes requested):

src/main/java/com/faforever/api/data/util/ElidePageSizeUtil.java
Ivan-Shaml: I agree, done. Although IMO, this is still a breaking change if the existing consumers expect to get 10K records.

Brutus5000 (approved):

Rogerstalker:
@Ivan-Shaml @Sheikah45 The ranking history only returns 100 elements now, so it looks like this:
kép

PR #1137, opened 2026-06-14
Allow exposing data to anonymous users as config option

Sheikah45:
Also add other security config changes to support development with gitops-stack tilt

## Release Notes

* **New Features**
  * Added `faf-api.allow-anonymous` configuration to control anonymous access.
  * When enabled, `/data/**` can be accessed without authentication.

* **Bug Fixes**
  * Improved role checking to safely handle unauthenticated/anonymous requests without errors.

* **Chores**
  * Updated local development configuration to enable anonymous access and use local Hydra endpoints for OAuth2 JWT (issuer/JWKS) settings.

### DIFF
--- a/src/main/java/com/faforever/api/config/FafApiProperties.java
+++ b/src/main/java/com/faforever/api/config/FafApiProperties.java
@@ -16,6 +16,7 @@ public class FafApiProperties {
    * The API version.
    */
   private String version;
+  private boolean allowAnonymous;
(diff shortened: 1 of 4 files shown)


Brutus5000:
Your changes overlap with https://github.com/FAForever/faf-java-api/pull/1138

But they do more, because the expose too much. Do you need anything more than in https://github.com/FAForever/faf-java-api/pull/1138 for Tilt?

